PR #135, opened 2016-07-06
Remove mnesia directory to fix filezcache boot failure

ddeboer:
See also https://github.com/zotonic/zotonic/issues/1321.


Makefile
```diff
@@ -44,6 +44,9 @@ up:
 	@echo "> Started. Open http://localhost in your browser."
 
 up-zotonic:
+	@export $(cat .env) > /dev/null
+	# See https://github.com/zotonic/zotonic/issues/1321
+	rm -rf $(ZOTONIC)/priv/mnesia/*
```
ddeboer: @mworrell Will this slow down the Zotonic boot procedure in any way?

mworrell: It will not slow down booting, but it will force deletion of all cached files.

ddeboer:
@Dirklectisch Did you test this? If so, we can merge it.

Issue #123, opened 2016-06-22
Hamburger menu UX


desteev:
Mobile devices: hamburger icon changes into a cross after opening the hamburger menu. To close the menu you can click the cross. The cross then changes back into a hamburger icon.

ddeboer:
Fixed in de8b5334112a5531bd55d4940e960464d0e854dd.

Issue #139, opened 2016-07-13
Live roadmap from GitHub


ddeboer:
Pull a live roadmap from the GitHub API so we can display that on ginger.nl.

PR #148, opened 2016-07-25
Add creating issues to guidelines

ddeboer:
Now we have a roadmap (#139), we need to make sure it shows all significant changes.

@Dirklectisch I understand you already had some discussion about this with @eminetuzkapan. This is my proposed solution.

### DIFF
--- a/docs/guidelines.md
+++ b/docs/guidelines.md
@@ -63,16 +63,23 @@ $ git merge release-0.5.0
 
 They will then become part of the next Ginger release.
 
-### 10. Prefer catinclude over include.
+### 10. Create an issue before doing any significant work.
+
+* This enables [others](https://about.gitlab.com/handbook/#gitlab-workflow) to 
+keep track of changes and provide you with ideas.
+
+* We tag issues to show them on the [roadmap](http://ginger.nl/roadmap).
+
+### 11. Prefer catinclude over include.
 
 `{% catinclude id %}` allows more specific template overrides.
 
-### 11. Place custom Erlang functionality in observers.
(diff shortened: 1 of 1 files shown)


Dirklectisch:
@ddeboer As I mentioned to @eminetuzkapan there will always be commits that do not result in any user facing changes. I am not sure how useful it will be to make explicit issues for those.

For stuff that does result in improvements for the user I totally agree with this approach.

Btw. Did you now you can also mention issues in commit messages by using the #123456 syntax? That will automatically associate commits with issues.

I also showed @eminetuzkapan the activity overviews on Github. It might be nice to incorporate this on our Ginger website so that users can see that development is indeed very active.

https://github.com/driebit/ginger/pulse
https://github.com/driebit/ginger/graphs/contributors

Just my two cents.

ddeboer:
> there will always be commits that do not result in any user facing changes.

True. I’m still in favour of having an issue for those as long as they involve ‘significant work’, i.e. work that takes some time and/or requires discussion to find the best solution.

We can always filter the issues shown on the roadmap by adding an (e.g.) ‘user’ label.

> Did you now you can also mention issues in commit messages by using the #123456 syntax? 

I did, of course. 😉  Not just mention but also close, by the way (`fix #123`).

> activity overviews on Github

Not sure end users will understand those graphs. @eminetuzkapan Please create a separate issue if we want to have this for our clients.

ddeboer:
By the way, my proposed solution is an alternative to showing all *commits* (instead of closed issues) on the roadmap – which will definitely confuse end users.

PR #151, opened 2016-08-01
Add 'make disco' command to enable discovery

CyBeRoni:
Enables foo.ginger.dev in the Driebit office lan. Since it is not dependent on the ginger containers running (and the ginger container isn't dependent on it), I figured it would be best to make it a separate command, to be run on-demand, which starts a container in daemon mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,3 +53,6 @@ up-zotonic:
 
 update:
 	@docker-compose pull
+
+disco:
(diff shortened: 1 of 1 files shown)


ddeboer:
Let’s add some documentation to `docker.md` as well.

ddeboer:
Thanks! 🔮

PR #152, opened 2016-08-02
Add haproxy container to mimic production environment.

CyBeRoni:
Add haproxy container to mimic production environment.

(Note: https redirect not active at this point.)


ddeboer:
This is not strictly necessary for the local development environment, so we’ll leave this out for now.

PR #155, opened 2016-08-09
Run module tests on Travis

ddeboer:


ddeboer:
Whoops, there are no mod_ginger_rdf tests as of yet. Let’s add some!

PR #159, opened 2016-08-11
added ignore for _sites. Needed to park unused sites because of docke…

gunnar-driebit:
added ignore for _sites. Needed to park unused sites because of docker performance.


ddeboer:
This is too specific. Please either:

- move the sites outside your Ginger dir
- add `_sites_` to your global `.gitignore`.

PR #185, opened 2016-09-19
better video handling for attached media

Devenwel:
@DorienD can you take a look please, play arrow styling is wrong and sometimes I am not getting a thumbnail when I embed via URL

### DIFF
--- a/modules/mod_ginger_base/templates/attached-media/attached-media-thumb.video.tpl
+++ b/modules/mod_ginger_base/templates/attached-media/attached-media-thumb.video.tpl
@@ -1,39 +1,72 @@
+{% if m.rdf[id.rdf] as rdf %}{% endif %}
+
 {% if
     id.medium.mime == "text/html-oembed" or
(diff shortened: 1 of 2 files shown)


modules/mod_ginger_base/templates/attached-media/attached-media-thumb.video.tpl
```diff
@@ -1,39 +1,72 @@
+{% if m.rdf[id.rdf] as rdf %}{% endif %}
```
ddeboer: Can be removed.

DorienD: Removed it

DorienD:
I have removed the empty rdf check and added some styling for the play icon in the css instead of the icon element.

ddeboer:
Is this ready to be merged?

DorienD:
@Devenwel This was done right?

Issue #170, opened 2016-09-05
PDF preview doesn't show right


kathleenvw:
When viewing a pdf in the preview on the bottom of the page (not as a pop-up/lightbox) it is to wide (it falls behind other content).
![ambities__doelstellingen_en_ruimtelijke_plannen_in_de_provincie_friesland_-_architectuur_lokaal](https://cloud.githubusercontent.com/assets/19573451/18244815/60109e88-7362-11e6-9645-c57cc2afa937.png)

For example see:
http://architectuurlokaal.ginger-test.driebit.net/page/9577/titel
http://architectuurlokaal.ginger-test.driebit.net/page/8289
http://nieuwwestexpress.ginger-test.driebit.net/nl/page/6895/nieuw-west-afgel-open

ddeboer:
@Devenwel Is this related to https://github.com/driebit/ginger/commit/3b3c5e23e33ec5293c2402f469167f9ad03a0a04?

gunnar-driebit:
I fixed this in the 170-pdf-preview-width branch.

ddeboer:
@gunnar-driebit Nice! Please open a PR for that branch and reference this issue, like this: `Fix #170`.

PR #192, opened 2016-09-28
170 pdf preview to fit in page

gunnar-driebit:
Fixed the problem with pdf preview being a fixed width, causing it to overflow its boundaries.


ddeboer:
Please add a reference, like this: `Fix #170`.

gunnar-driebit:
Fix #170

ddeboer:
@Devenwel Do you agree with this fix? It applies to your code.

Devenwel:
Can't remember why I made it 800 px wide, but if the issue is fixed, fine with me.

PR #194, opened 2016-09-29
191 masthead image without cropping

### LINKED ISSUE #191: High resolution masthead banner should not be cropped
High resolution banners seem to be cropped by default, and because of the cropping, the image does not show in high res anymore. 
Found out about this at tudelft.openresearch.net and reproduced here: https://mijnzuiderzee.ginger-acceptatie.driebit.net/page/6579/test-zoekopdracht (the image I uploaded is 2900 x 700 px).

gunnar-driebit:
Fix #191

- [x] Dorien removes inline styling
- [x] Dorien checks mediaclass config if only {quality, 85%} is enough

### DIFF
--- a/modules/mod_ginger_base/templates/masthead/masthead.tpl
+++ b/modules/mod_ginger_base/templates/masthead/masthead.tpl
@@ -22,11 +22,11 @@ as
         {% else %}
 
             {% if id.o.hasbanner[1].depiction.width > 500 %}
-                <div class="masthead do_parallax {{ extraClasses }}" style="background: transparent url({% image_url id.o.hasbanner[1].depiction.id lossless %}) no-repeat center center; background-size: cover;"></div>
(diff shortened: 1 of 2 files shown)


fredpook:
I think the mediaclass should stay since we want to downscale very large images and upscale low res images. Can you test with adjusting mediaclass.config ?

gunnar-driebit:
I changed the mediaclass.config file in mod_ginger_foundation to optimize the quality vs file size.

ddeboer:
@fredpook Can you test this branch and see how the masthead images turn out?

ddeboer:
Ping @fredpook.

DorienD:
@fredpook allright, i've removed all the inline styling that was not necessary and updated the mediaclass config so it doesn't crop the images. Can you merge this?

PR #197, opened 2016-09-30
Added button

DorienD:
I've added a button so it is visual that you can click on the embed block. Finn, can you help me next week with the styling of this button?


ddeboer:
@finnkruyning Please look into this.

PR #199, opened 2016-10-05
Toggle definitions in body text

DorienD:
I've added the javascript and the styles for this. One thing remains and that is the configuration of the tinymce. I'm not sure where to put this configuration. @ddeboer what do you think? Make a default config in ginger base or in ginger foundation? Now the admin_tinymce overrides is in the ginger edit module. Not sure what will happen if we also put a file in base or foundation, @fredpook you have some experience with this?


ddeboer:
Can you move `_admin_tinymce_overrides_js.tpl` to ginger_base? After all, the config applies not only to (frontend) edit but also to the admin.

ddeboer:
Ping @DorienD.

ddeboer:
So was this resolved?

PR #216, opened 2016-11-09
[rdf] Include JSON-LD in pages

### LINKED ISSUE #179: Link to JSON-LD
Google’s structured data testing tool indicates it [can process](https://search.google.com/structured-data/testing-tool?hl=nl#url=https%3A%2F%2Fwww.joodsmonument.nl%2Frdf%2F557703) our JSON-LD documents. However, the documents are not yet discovered by Google.

I’d think something like`<link rel="alternate" type="application/ld+json" href="https://www.ginger.nl/rdf/123">` should work but there’s some [discussion](http://webmasters.stackexchange.com/questions/83372/why-dont-search-engines-support-json-ld-schema-org-as-seperated-endpoints-via) as to whether this is possible at all or Google parses inline JSON-LD only.

ddeboer:
Fix #179.

* Add an alternate link.
* Because search engines don't follow those links, also include the JSON-LD in the page.
* Tested against Google's structured data testing tool (https://search.google.com/structured-data/testing-tool).

### DIFF
--- a/modules/mod_ginger_rdf/models/m_rdf.erl
+++ b/modules/mod_ginger_rdf/models/m_rdf.erl
@@ -20,6 +20,8 @@
 
 m_find_value(#rdf_resource{} = Rdf, #m{value = undefined} = M, _Context) ->
     M#m{value = Rdf};
+m_find_value(Id, #m{value = undefined}, Context) when is_integer(Id) ->
+    mochijson2:encode(ginger_json_ld:serialize(to_triples(Id, Context)));
 m_find_value(id, #m{value = #rdf_resource{id = Id}}, _Context) ->
     Id;
 m_find_value(uri, #m{value = #rdf_resource{id = Id}}, _Context) ->
--- a/modules/mod_ginger_rdf/templates/_html_head.tpl
+++ b/modules/mod_ginger_rdf/templates/_html_head.tpl
@@ -1,0 +1,1 @@
+    <link rel="alternate" type="application/ld+json" href="{% url rsc_json_ld id=id %}">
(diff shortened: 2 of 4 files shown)


ddeboer:
@Dirklectisch Please review and merge if ok.

modules/mod_ginger_rdf/models/m_rdf.erl
```diff
@@ -20,6 +20,8 @@
 
 m_find_value(#rdf_resource{} = Rdf, #m{value = undefined} = M, _Context) ->
     M#m{value = Rdf};
+m_find_value(Id, #m{value = undefined}, Context) when is_integer(Id) ->
+    mochijson2:encode(ginger_json_ld:serialize(to_triples(Id, Context)));
```
Dirklectisch: @ddeboer I think this will be a problem for when you are using a model to chain a query. For example m.rdf[id].predicate will not work anymore? You could use the to_value function to realise the actual value and return and #m record here instead.

ddeboer: But `m_value` is only triggered for `{{ m.rdf }}`, not `{{ m.rdf[id }}`.

Anyway, there shouldn’t be a problem, because of the `is_integer` guard. You 
- either pass an RDF resource (`m.rdf[id.rdf]['http://bla.com/some_predicate']`)
- or call `m.rdf[id]` with a rsc id and without any arguments, which will dump the whole thing.

Dirklectisch: Sorry missed the guard. Tested the branch locally on JHM seems to work. I'll merge it now.

BTW. I do think m_value is called for when you do `{{ m.rdf[id] }}` if you return an m# record.

ddeboer: Good point. Enjoy dd7cc1eae87a876863a30cc2a8b6e660dc8dfee3. :wink

PR #217, opened 2016-11-10
Fix ginger edit connection new rsc

loetie:
Removed callback from new rsc template since there is now a live template to show the connected resources. This callback was from a previous version.

### DIFF
--- a/modules/mod_ginger_edit/templates/_action_ginger_dialog_new_rsc_tab.tpl
+++ b/modules/mod_ginger_edit/templates/_action_ginger_dialog_new_rsc_tab.tpl
@@ -1,5 +1,5 @@
 {% with redirect|default:(dispatch=="ginger_edit")|if:[]:`ginger_edit` as redirect %}
-{% with callback|default:(dispatch=="ginger_edit")|if:"zAdminConnectDone":"" as callback %}
+{% with callback|default:"" as callback %}
 {% with actions|default:[] as actions %}
 
 


modules/mod_ginger_edit/templates/_action_ginger_dialog_new_rsc_tab.tpl
```diff
@@ -1,5 +1,5 @@
 {% with redirect|default:(dispatch=="ginger_edit")|if:[]:`ginger_edit` as redirect %}
-{% with callback|default:(dispatch=="ginger_edit")|if:"zAdminConnectDone":"" as callback %}
+{% with callback|default:"" as callback %}
```
ddeboer: Is this backwards compatible in the situation where no `callback` argument is passed (which before defaulted to `ginger_edit`)?

loetie: Fred+1.

Before it was defaulted to "zAdminConnectDone" if it was from a "ginger_edit" page. Otherwise it was "". Since there is a live reload within ginger_edit there is no need for "zAdminConnectDone" to be executed here, so default "". Custom callbacks should keep working, and ginger_edit is changed. Only if sites still use the old version it could be problem?

fredpook: please, test on jhm_tentoonstellingen

ddeboer: Ah, already merged. But please test. We can always revert this if it causes problems. :wink:

PR #221, opened 2016-11-11
404

### LINKED ISSUE #212: editable 404 page
Would be nice to have the option to edit your 404. Clients can make there own 404 page with blocks (text, media etc).

DorienD:
Added unique page_404 to ginger base schema so 404 pages can now be filled in by content the user specifies. 

Fix #212.

### DIFF
--- a/modules/mod_ginger_base/mod_ginger_base.erl
+++ b/modules/mod_ginger_base/mod_ginger_base.erl
@@ -8,7 +8,7 @@
 -mod_description("Ginger Base").
 -mod_prio(250).
 -mod_depends([mod_content_groups, mod_acl_user_groups]).
--mod_schema(8).
+-mod_schema(9).
 
 -export([
     init/1,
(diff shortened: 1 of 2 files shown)


modules/mod_ginger_base/mod_ginger_base.erl
```diff
@@ -8,7 +8,7 @@
 -mod_description("Ginger Base").
 -mod_prio(250).
 -mod_depends([mod_content_groups, mod_acl_user_groups]).
--mod_schema(7).
+-mod_schema(9).
```
ddeboer: Why not 8? :wink:

DorienD: Good question, i think i got my version screwed and chose 9 instead of 8. We could bring this back to 8.

ddeboer: It’s not really a problem, though, so just leave it at 9.

Dirklectisch: @DorienD He's pulling your leg. Module downgrades are prohibited so you end up bumping the version number several times during development unless you want to mess with your DB.

PR #222, opened 2016-11-11
Banner

### LINKED ISSUE #214: Warning bar in acceptance environment
It happens sometimes that clients are using the acceptance environment when thinking that they are at production/live.  To prevent mistakes it would be good to see a clear warning in Ginger (YOU ARE WORKING IN ACCEPTANCE). Is it possible to have a standard warning bar (like a cookie bar) in all sites that reminds you of this. (maybe even possible at the frontend of the site too).

DorienD:
Fix #214.

- [x] Add unpublished banner rsc in schema
- [x] Change template check to look for (published!) banner rsc as well
- [ ] Add a close button
- [ ] Add a cookie on the close button

### DIFF
--- a/modules/mod_ginger_banner/mod_ginger_banner.erl
+++ b/modules/mod_ginger_banner/mod_ginger_banner.erl
@@ -1,0 +1,25 @@
+-module(mod_ginger_banner).
+-author("Driebit <info@driebit.nl>").
+
+-mod_title("Ginger banner module").
+-mod_description("Shows a banner on top of the site, if this is added in the config of the site or trough the rsc banner.").
+-mod_prio(500).
+
+-mod_schema(1).
+
+-include_lib("zotonic.hrl").
+
+-export([
+    manage_schema/2
+]).
+
+manage_schema(_Version, Context) ->
+    Datamodel = #datamodel{
+        resources = [
+            {message_banner, text, [
+                {title, "Banner"},
+                {is_published, false}
+            ]}
+        ]
+    },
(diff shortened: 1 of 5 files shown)


DorienD:
@ddeboer i need some help with the prio of this module. I have it on 500 now but it stills adds the _html_head.tpl after the base _html_head.tpl. Which is giving conflicts with my styles.

modules/mod_ginger_banner/mod_ginger_banner.erl
```diff
@@ -0,0 +1,8 @@
+-module(mod_ginger_banner).
+-author("Driebit <info@driebit.nl>").
+
+-mod_title("Ginger banner module").
+-mod_description("Shows a banner on top of the site, if this is added in the config of the site or trough the rsc banner.").
+-mod_prio(500).
```
ddeboer: Should be a *lower* number than mod_ginger_base (which is 250): see also the [docs](http://docs.zotonic.com/en/latest/developer-guide/modules.html#priority).

ddeboer:
Yep, would be nice, but is not strictly necessary. Let’s start with this simple version and see how editors use it before we start adding more features.

DorienD:
@ddeboer can you merge this, if you think this is good.

ddeboer:
Missing a readme: see [guidelines](https://github.com/driebit/ginger/blob/master/docs/guidelines.md#14-create-useful-ginger-modules).

DorienD:
Readme added! Now the only thing that remains is setting a cookie and a close button so you can click it away.

PR #226, opened 2016-11-14
Summary

### LINKED ISSUE #223: Use summary filter where appropriate
For instance, replace [this](https://github.com/driebit/ginger/blob/master/modules/mod_ginger_base/templates/list/list-item-vertical.tpl#L59-L63) with: `id|summary:70`.

DorienD:
#223, ive changed it for the list items, couldn't find it elsewhere. @ddeboer @fredpook do you know where its also used? Fix #223

### DIFF
--- a/modules/mod_ginger_base/templates/list/list-item-vertical.tpl
+++ b/modules/mod_ginger_base/templates/list/list-item-vertical.tpl
@@ -56,18 +56,10 @@
                     {% if id.subtitle %}
                         <p><b>{{ id.subtitle|striptags }}</b></p>
                         {% if id.subtitle|length<35 %}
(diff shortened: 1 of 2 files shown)


ddeboer:
If you do a search for `id.summary`, you’ll find some more places, but they don’t always have the same logic.

ddeboer:
Please add `Fix #223` to the PR description, so the issue will be automatically closed when we merge this.

DorienD:
Yes indeed, i'm also not sure if we can just add it everywhere. @fredpook could you also take a look at this?

ddeboer:
Not in the title, but in the description. :)

PR #236, opened 2016-11-24
[base] Make Ginger search open for extension

ddeboer:
* Replace call to search_query:search with notification.
* Make sure to remove all custom search arguments, as Zotonic
  will trip over them.

### DIFF
--- a/modules/mod_ginger_base/support/ginger_search.erl
+++ b/modules/mod_ginger_base/support/ginger_search.erl
@@ -2,6 +2,8 @@
 
 -export([
     search_query/2,
+    search_sql/2,
+    query_arguments/2,
     get_unfindable_categories/1,
     merge_ginger_args/2,
     withdefault/2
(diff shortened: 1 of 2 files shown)


modules/mod_ginger_base/support/ginger_search.erl
```diff
@@ -10,20 +10,37 @@
...
+            lists:append([Args, [{filters, Filters}]])
     end,
 
     QueryArgs = merge_ginger_args(Args1, Context),
-    search_query:search(QueryArgs, Context).
+    Query = #search_query{search = {'query', QueryArgs}, offsetlimit = Offset},
+
+    %% Forward to the next observer. Make sure all custom Ginger search
+    %% arguments have been removed.
+    z_notifier:first(Query, Context).
```
ddeboer: @Dirklectisch Was there any compelling reason *not* to do this in the first place? After removing the custom search arguments, forwarding to Zotonic’s search (or in my case, Elasticsearch) seems to work well.

modules/mod_ginger_base/support/ginger_search.erl
```diff
@@ -10,20 +10,37 @@
...
         undefined ->
-            Args1 = Args;
+            Args;
         Filters ->
-            Args1 = lists:append([Args, [{filters, Filters}]])
+            lists:append([Args, [{filters, Filters}]])
     end,
 
     QueryArgs = merge_ginger_args(Args1, Context),
-    search_query:search(QueryArgs, Context).
```
ddeboer: This blocked the chain of observers, which is almost always a bad thing as it makes the code [closed for extension](https://en.wikipedia.org/wiki/Open/closed_principle).

ddeboer:
Adapted with a quick solution to remain compatible with ginger_geo_search. I considered refactoring of the Zotonic search notifications (split `search_query` into two notifications, as discussed verbally) out of scope for this PR.

Dirklectisch:
I tested this on one of our sites that has a map on the homepage and it seems to work fine.

PR #195, opened 2016-09-29
[remark] Don't link to the anonymous user

fredpook:
@DorienD can you check if I did it right?

PR #241, opened 2016-11-30
[remark] Fix e-mail link missing for anonymous user

ddeboer:
* This was broken in #195.
* This PR drops the fallback to creator.
* Remain compatible with anonymous remarks imported from Anymeta.


ddeboer:
Updated with a check on `human_user` for compatibility with Anymeta.

PR #247, opened 2016-12-16
Masthead checks parent masthead image

DorienD:
A check if the parent or the rsc has a banner or depiction to use as a header


DorienD:
I've also added a check on crop centering for vertical centering of the mastehad

PR #261, opened 2017-02-03
Importing a db should not stop zotonic but only the site

Dirklectisch:
I think this is a little less frustrating than having to restart and rebuild all the containers on every DB import.


ddeboer:
Good one, thanks!

PR #267, opened 2017-03-02
[remark] Do not load TinyMCE JS on all pages

### LINKED ISSUE #266: Move _editor.tpl out of script from ginger remark
When ginger remark is activated the tinymce javascript is included on every page. But you only want to include those scripts on the pages that actually have remarks.

ddeboer:
Fix #266.

* Remove from _script.tpl, as it's loaded on all pages.
* Move overrides to custom template.


fredpook:
Works like a charm!

PR #268, opened 2017-03-03
[base] search word in title on search page

DorienD:
Is this ok? "Search | seach word" or do you want "search word | Search"?

### DIFF
--- a/modules/mod_ginger_base/templates/search.tpl
+++ b/modules/mod_ginger_base/templates/search.tpl
@@ -1,6 +1,6 @@
 {% extends "base.tpl" %}
 
-{% block title %}{_ Search _}{% endblock %}
+{% block title %}{_ Search _} | {{ q.qs }}{% endblock %}
 
 {% block body_class %}t--search{% endblock %}
 


modules/mod_ginger_base/templates/search.tpl
```diff
@@ -1,6 +1,6 @@
 {% extends "base.tpl" %}
 
-{% block title %}{_ Search _}{% endblock %}
+{% block title %}{_ Search _} | {{ q.qs }}{% endblock %}
```
ddeboer: What about `Search: bli bla`? Reads more naturally.

krista-ramakers:
I agree, 'Search: search query' looks good to me!

ddeboer:
Ping @DorienD.

ddeboer:
Why merge this when the comments haven’t been taken into account?

DorienD:
Now it is search: q.qs so that according to the comments right?

ddeboer:
[Nope](https://github.com/driebit/ginger/commit/dc524a1f864aed55565e4b9f19b35b05877fad8c).

PR #274, opened 2017-03-23
Beeldenzoeker back to top button fix

yorivanloo:
Fixed infinite scroll and back to top button visibility


modules/mod_ginger_collection/lib/js/beeldenzoeker.js
```diff
@@ -56,13 +56,21 @@ $(document).ready(function(){
 
 
 function homeInfinite() {
+	staticWindowHeight = $(window).height();
```
ddeboer: Spaces, not tabs, as described in our [.editorconfig](https://github.com/driebit/ginger/blob/master/.editorconfig). Please install an Editorconfig plugin for your IDE.

ddeboer:
Replaced by #275.

PR #275, opened 2017-04-04
[base] Add back to top

ddeboer:
Make back to top functionality, originally developed for mod_ginger_collection, generic and move it into base/foundation. 

Replaces #274.


ddeboer:
@DorienD Ok with you? This should be approach to Ginger in general: when possible, make it generic. Back to top is generic functionality, so should be in base/foundation and not tied in with a specific homepage (e.g. the mod_ginger_collection homepage).

PR #279, opened 2017-04-18
Add automated browser tests

ddeboer:
This PR adds an automated browser testing setup POC.

1. This setup is based on JavaScript, as that makes it independent from the backend stack (Erlang for Ginger sites, PHP for some other projects, external APIs for SPAs). JavaScript is the common denominator between our projects. Moreover, it’s the browser’s language.

2. I opt for Selenium WebDriver instead of a headless NodeJS browser (such as PhantomJS). While PhantomJS will execute tests faster, it makes them fail randomly and thus raise tests maintenance costs. Later, when Chrome headless is stable, we should be able to migrate to that relatively easily. 

3. Tests must work both locally (with a view on the automated browser) and on CI for validation.

4. Tests will be stored in each site’s repository, so they’ll be visible to developers and fail as quickly as possible after the commit that caused the error was pushed.

5. These tests can rely on a base testing framework, which will be part of Ginger: a docker-compose config, Makefile and a package.json with the tools of our testing trade. This package.json can eventually be published as a separate NPM package, though that introduces some version management overhead. 

To do:

- [ ] documentation
- [x] choose between Gherkin (high-level) or plain JS (lower-level) as testing language
- [x] choose between (a combination of) [Cucumber](https://github.com/cucumber/cucumber-js) (BDD)/Mocha/Jasmine/Nightwatch as testing framework
- [x] add some tests to one of our sites
- [ ] how to trigger tests based on Zotonic/Ginger releases (see also 4 above)? Commits in those projects can cause test failures as well. We could trigger builds through the GitLab API and specify the Ginger and Zotonic version to test against in environment variables.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ data/*
 sites/*
 modules/*
 !modules/mod_ginger*
+tests/node_modules
+tests/reports
+tests/screenshots
 *.mo
 *.pot
 .idea
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ help:
 	@echo "  import-db-backup       Import database from a backup (host=ginger.driebit.net site=site-name)"
 	@echo "  shell                  Open Zotonic shell"
 	@echo "  psql                   Open PostgreSQL interactive terminal"
+	@echo "  test site=your_site	Run brower site tests"
 	@echo "  up                     Start containers"
 	@echo "  up-zotonic             Start containers with custom Zotonic checkout"
 	@echo "  update                 Update containers"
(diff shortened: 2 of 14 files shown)


ddeboer:
For illustration purposes, in Mocha, a test looks like this:

```javascript
before(() => driver.navigate().to('http://path.to.test.app/'));

it('autocompletes the name field', function() {
    driver.findElement(By.css('.autocomplete')).sendKeys('John');
    driver.wait(until.elementLocated(By.css('.suggestion')));
    return driver.findElement(By.css('.suggestion')).click();
});

after(() => driver.quit());
```

In CucumberJS:

```
Feature: Example feature
  As a user of Cucumber.js
  I want to have documentation on Cucumber
  So that I can concentrate on building awesome applications

  Scenario: Reading documentation
    Given I am on the Cucumber.js GitHub repository
    When I click on "CLI"
    Then I should see "Running specific features"
```

See also https://watirmelon.blog/2015/12/08/comparison-of-javascript-browser-automation-and-test-specification-libraries/

ddeboer:
Decided to go with CucumberJS on NightwatchJS.

PR #280, opened 2017-04-18
[collection] image not clickable fix

yorivanloo:


DorienD:
Did you test this in multiple browsers? Also IE?

yorivanloo:
Yes

PR #282, opened 2017-04-18
removed es6 functionality (ipad3 crashes)

yorivanloo:


ddeboer:
Can’t we use Babel, so we can move forward and start using ES6 syntax?

Devenwel:
sure, I have a Gulp task in entoen.nu that does this if you need an example

Devenwel:
Also the fat arrows change the way 'this' is bound so be careful in any case rewriting that stuff

ddeboer:
So @Devenwel, since we decided not to use ES6 syntax for now, should we merge this?

PR #285, opened 2017-04-25
Create separate ginger_embed module

DorienD:
Ginger embed module for embedding ginger resources on other ginger sites or not ginger sites


ddeboer:
What do we gain by the separate module, apart from mod_ginger_rdf? Mod_ginger_rdf still needs to be enabled for the embeds to work.

DorienD:
When it was made for hartamsterdam the files were all over te place, now its centralised in one module. And yes rdf still needs to be enabled. Hmmm.

DorienD:
I could move everything into mod_ginger_rdf. Also the erlang. Should we then make a config key if you want to have the ginger embeds or not?

modules/mod_ginger_embed/templates/embed/share-link.tpl
```diff
@@ -0,0 +1,2 @@
+<a id="{{ #embed }}" href="" title="{_ Embed _}" class="">{_ Embed _}</a>
```
ddeboer: This file has the wrong path: it must be `templates/share/share-link.tpl` to be picked up by mod_ginger_base’s `share.tpl`.

ddeboer:
>  Should we then make a config key if you want to have the ginger embeds or not?

Actually, when there’s something to toggle (enable/disable), a separate module makes sense: enable/disable the module.

In this case the only toggable is the `share-link.tpl` template, isn’t it?

modules/mod_ginger_embed/mod_ginger_embed.erl
```diff
@@ -0,0 +1,62 @@
+%% @author Driebit <info@driebit.nl>
+%% @copyright 2015
+
+-module(mod_ginger_embed).
+-author("Driebit <info@driebit.nl>").
+
+-mod_title("Ginger Embed").
+-mod_description("Ginger Embed").
+-mod_prio(250).
+-mod_depends([mod_content_groups, mod_acl_user_groups]).
```
ddeboer: Add `mod_ginger_rdf`.

DorienD: done.

DorienD:
I've moved the share link, so it works together with the all include from ginger base.

PR #289, opened 2017-05-02
[collection] fixed filter collapse bug on certain filters

yorivanloo:


yorivanloo:
@ddeboer not sure if this one is still applicable

ddeboer:
No longer relevant as separate filters have been replaced with one generic type: `filter-terms`.

PR #290, opened 2017-05-02
[collection] added new filter functionality to the search filter load more instead…

yorivanloo:
… of scroll


modules/mod_ginger_collection/lib/js/search/components/filters-subjects.js
```diff
@@ -8,11 +8,33 @@ $.widget('ui.search_cmp_filters_subjects', {
...
+                if(i < counter) {
+                    $(val).removeClass('hidden');
+                }
+            });
+
+            if(counter + counter > listItem.length) {
+                $(this).hide();
+            }
+        });
+    },
```
ddeboer: This makes the ‘load more’ functionality dependent on this particular filter. That seems wrong to me: it’s generic functionality that is applicable to all filters.

PR #294, opened 2017-05-16
Restructure config override files so it can be overwritten on site level

loetie:
Currently it it impossible to override tinymce templates within site level. The override templates are all included with the editor.

The all include loads the site template first, then the module template. So the options in the module template can never be changed by the site template.

The idea is that we can set a parameter on site level and check that on module level to see if the module should add configuration or not.

### DIFF
--- a/modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
+++ b/modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
@@ -1,7 +1,19 @@
 /*
 Custom settings to override tiny-init.js.
 */
-tinyInit.language="{{ m.translation.language|default:"en" }}";
-tinyInit.menubar="";
-tinyInit.toolbar="styleselect | bold italic | bullist numlist | removeformat | zlink zmedia | link unlink | code";
-tinyInit.extended_valid_elements="iframe[src|style|width|height|scrolling|marginwidth|marginheight|frameborder]";
+
+if (typeof tinymce_overrides_language === 'undefined' || !tinymce_overrides_language) {
+    tinyInit.language="{{ m.translation.language|default:"en" }}";
+}
+
+if (typeof tinymce_overrides_menubar === 'undefined' || !tinymce_overrides_menubar) {
+    tinyInit.menubar="";
+}
+
+if (typeof tinymce_overrides_toolbar === 'undefined' || !tinymce_overrides_toolbar) {
+    tinyInit.toolbar="styleselect | bold italic | bullist numlist | removeformat | zlink zmedia | link unlink | code";
+}
+
+if (typeof tinymce_overrides_extended_valid_elements === 'undefined' || !tinymce_overrides_extended_valid_elements) {
+    tinyInit.extended_valid_elements="iframe[src|style|width|height|scrolling|marginwidth|marginheight|frameborder]";
+}
--- a/modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
+++ b/modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
@@ -2,7 +2,7 @@
 Custom settings to override tiny-init.js.
 */
 
-if (typeof tinymce_overrides_language === 'undefined' || !tinymce_overrides_language) {
+if (typeof tinyInit.language === 'undefined') {
     tinyInit.language="{{ m.translation.language|default:"en" }}";
 }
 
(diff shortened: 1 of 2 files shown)


modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
```diff
@@ -1,7 +1,19 @@
...
 Custom settings to override tiny-init.js.
 */
-tinyInit.language="{{ m.translation.language|default:"en" }}";
-tinyInit.menubar="";
-tinyInit.toolbar="styleselect | bold italic | bullist numlist | removeformat | zlink zmedia | link unlink | code";
-tinyInit.extended_valid_elements="iframe[src|style|width|height|scrolling|marginwidth|marginheight|frameborder]";
+
+if (typeof tinymce_overrides_language === 'undefined' || !tinymce_overrides_language) {
+    tinyInit.language="{{ m.translation.language|default:"en" }}";
+}
```
ddeboer: Why not just check if `tinyInit.language` itself is already set (instead of introducing a separate variable, which requires clients to set that variable, too)?

modules/mod_ginger_remark/templates/_tinymce_overrides.tpl
```diff
@@ -1,17 +1,21 @@
...
-                                {title: "Underline", icon: "underline", format: "underline"},
-                                {title: "Strikethrough", icon: "strikethrough", format: "strikethrough"},
-                            ]},
-                            {title: "Blocks", items: [
-                                {title: "Paragraph", format: "p"},
-                                {title: "Blockquote", format: "blockquote"},
-                            ]}
-                        ];
+if (typeof tinymce_overrides_style_formats === 'undefined' || !tinymce_overrides_style_formats) {
+    tinyInit.style_formats = [  {title: "Headers", items: [
```
ddeboer: Here the same as for `language` can be done: check `tinyInit.style_formats` itself.

modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
```diff
@@ -1,7 +1,19 @@
...
+if (typeof tinymce_overrides_menubar === 'undefined' || !tinymce_overrides_menubar) {
+    tinyInit.menubar="";
+}
+
+if (typeof tinymce_overrides_toolbar === 'undefined' || !tinymce_overrides_toolbar) {
+    tinyInit.toolbar="styleselect | bold italic | bullist numlist | removeformat | zlink zmedia | link unlink | code";
+}
+
+if (typeof tinymce_overrides_extended_valid_elements === 'undefined' || !tinymce_overrides_extended_valid_elements) {
+    tinyInit.extended_valid_elements="iframe[src|style|width|height|scrolling|marginwidth|marginheight|frameborder]";
```
ddeboer: Just use `+=` here to append `extended_valid_elements` instead of overwriting them.

modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
loetie: That seperate parameter is needed because most tinyInit options will always return true because they have already been set in this javascript file which is include just before the override templates.

https://github.com/zotonic/zotonic/blob/master/modules/mod_editor_tinymce/lib/js/tinymce-4.5.5/tiny-init.js

modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
loetie: We could do this, but then it would not be possible to remove an element from the allowed list. We could use a parameter to determine if it should overwrite or add?

modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
ddeboer: But `language` and `style_formats` (see below) haven’t.

modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
ddeboer: `extended_valid_elements` is not set by Zotonic, so empty by default. By changing to `+=` here any values set by the site will be kept (not removed).

modules/mod_ginger_remark/templates/_tinymce_overrides.tpl
loetie: Done

modules/mod_ginger_edit/templates/_admin_tinymce_overrides_js.tpl
loetie: Done. But now it is not possible anymore to remove valid element defined by the module.

PR #298, opened 2017-05-23
[collection] seperated meta-data links with new meta-link.tpl

yorivanloo:


ddeboer:
What problem does this PR solve? Please add a description to clarify that.

PR #299, opened 2017-05-29
[collection] added book icon + image icon with text

yorivanloo:


ddeboer:
The book icon (and the collection names) is specific to Zuiderzee, so should not be in Ginger.

PR #305, opened 2017-06-26
[WIP] [collection] Add linked data blocks

ddeboer:


ddeboer:
Added in a separate change set.

PR #306, opened 2017-06-29
[Collection] Added features and improved design

rl-king:


modules/mod_ginger_collection/lib/css/src/blocks/_masthead.scss
```diff
@@ -5,12 +5,16 @@
 	top: 0px;
 	left: 0px;
 	overflow: hidden;
-	
-	&.no-image {
+
+	.no-image {
```
ddeboer: This looks like a mistake: `<div class="masthead no-image"></div>`.

PR #309, opened 2017-07-05
[collection] Object page styling

rl-king:
Please check if this is the correct solution to conditionally rendering the metadata element.
https://github.com/driebit/ginger/commit/c6fee9b2fe1f84cec49eba0be00fdc82699e7cc3

### DIFF
--- a/modules/mod_ginger_collection/templates/list/list-item-beeldenzoeker.tpl
+++ b/modules/mod_ginger_collection/templates/list/list-item-beeldenzoeker.tpl
@@ -7,13 +7,15 @@
 
             <div class="list__item__content">
                 {% block item_content %}
-                    {% block item_meta %}
-                        <div class="list__item__content__meta">
-                            {% if record['production.date.start'] %}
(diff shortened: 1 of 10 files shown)


modules/mod_ginger_collection/templates/list/list-item-beeldenzoeker.tpl
```diff
@@ -7,13 +7,15 @@
...
             <div class="list__item__content">
                 {% block item_content %}
-                    {% block item_meta %}
-                        <div class="list__item__content__meta">
-                            {% if record['production.date.start'] %}
-                                <time datetime="{{ record['production.date.start'] }}">{{ record['production.date.start'] }}{% if record['production.date.end'] and record['production.date.end'] != record['production.date.start'] %}&#8202;–&#8202;{{ record['production.date.end'] }}{% endif %}</time>
-                            {% endif %}
-                        </div>
-                    {% endblock %}
+                    {% if item_meta %}
```
ddeboer: Where is the `item_meta` passed into this template?

PR #310, opened 2017-07-06
Tinymce default languages

fredpook:
Setting the admin language switch to another language then nl/en/ru causes a 404 on
/lib/js/tinymce-4.3.7/tinymce/plugins/langs/##.js
there are only 3 language javascripts


fredpook:
moved to https://github.com/driebit/ginger/tree/tinymce_default-lang2 and closing this

PR #326, opened 2017-08-25
New Gulp setup

rl-king:
This PR updates the the Ginger skeleton used when creating a new site at `make addsite`

**Added features**
- Livereload scss, js, and tpl files
- Transpiles javascript to the last two browser versions
- Minify javascript

It also adds the two default `/lib/css/src/screen.scss` and `/lib/js/src/script.js` files.

The package.json has been updated with 
```    
    "name": "ginger",
    "version": "1.0.0",
    "description": "Ginger skeleton",
    "scripts": {
        "test": "echo \"Error: no test specified\" && exit 1",
        "start": "gulp",
        "build": "gulp sass"
    },
    "author": "Driebit",
    "license": "UNLICENSED",
```

And the `dependencies` renamed to `devDependencies`

### DIFF
--- a/skel/gulpfile.js
+++ b/skel/gulpfile.js
@@ -1,44 +1,82 @@
-'use strict';
+const gulp = require('gulp');
 
-var gulp = require('gulp'),
-    sass = require('gulp-sass'),
-    postcss = require('gulp-postcss'),
-    autoprefixer = require('autoprefixer'),
-    lost = require('lost'),
-    globbing = require('gulp-css-globbing'),
-    livereload = require('gulp-livereload');
+// Scss
+const sass = require('gulp-sass');
+const postcss = require('gulp-postcss');
+const autoprefixer = require('autoprefixer');
+const lost = require('lost');
+const globbing = require('gulp-css-globbing');
 
-var paths = {
-    modulesSrc: [
-        '../../modules/mod_ginger_foundation/lib/css/src',
-    ],
-    cssSource: 'lib/css/src/',
-    cssDestination: 'lib/css/site/'
+// Javascript
+const babel = require('gulp-babel');
+const uglify = require('gulp-uglify');
+const rename = require('gulp-rename');
+
+// Prevents the pipe/stream from breaking on error
+const plumber = require('gulp-plumber');
+
+// Livereload
+const livereload = require('gulp-livereload');
+
+// Locations of scss, javascript and template files
+const paths = {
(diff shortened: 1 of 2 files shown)


skel/gulpfile.js
```diff
@@ -1,44 +1,83 @@
...
-        paths.cssSource + '/**/*.scss'
-    ].concat(modules);
+    var scssCombinedPaths = paths.modulesSrc
+        .map(x => `${x}/**/*.scss`)
+        .concat(`${paths.scssSrc}/**/*.scss`);
 
-    gulp.watch(watchPaths, ['sass']);
+    gulp.watch(scssCombinedPaths, { interval: 100 }, ['sass']);
+    gulp.watch(`${paths.templateSrc}/**/*.tpl`, { interval: 100 }, ['tpl']);
+    gulp.watch(`${paths.javascriptSrc}/**/*.js`, { interval: 100 }, ['js']);
```
ddeboer: Does `interval` change the behaviour? I _think_ it only matters for polling, we we don’t (or at least shouldn’t) do: we have inotify.

rl-king: It's indeed the polling interval which I had setup to reduce CPU usage. It's something I tested years ago, something that might not be relevant anymore today.

ddeboer: Can you test whether it makes any difference? My hypothesis is that it shouldn’t.

rl-king:
Hold on with merging this PR. Babel doesn't compile certain new JS features to IE11 compatible code apparently. We might need to include a polyfill for this, like https://babeljs.io/docs/usage/polyfill/. I'll do some more research, adding a runtime dependancy by default is not ideal.

skel/gulpfile.js
rl-king: There is no significant difference in cpu usage by watching multiple paths simultaneously.

ddeboer:
The polyfill was added in #330. @rl-king Is this now ready for merging?

skel/gulpfile.js
ddeboer: And about the polling: as it turns out, the interval *does* matter. Gulp 3 *polls* for changes. 😞 Things seem to be better in Gulp 4 (unreleased) and tools like Webpack.

rl-king:
This can be merged. I've removed the interval from the watcher and a specific babel compilation targets, which seemed unnecessary.

ddeboer:
Thanks!

PR #328, opened 2017-08-28
[collection] Set fixed height to map

yorivanloo:
This solves the problem: map block overflows block underneath.


modules/mod_ginger_collection/templates/collection/metadata/geo.tpl
```diff
@@ -6,7 +6,7 @@
         </h6>
         <div class="adlib-object__meta__data">
             {% include "list/list-uri-labels.tpl" items=places %}
-            {% include "map/map.tpl" result=places %}
+            {% include "map/map.tpl" result=places height=200 %}
```
ddeboer: Doesn’t this change the height from 300 to 200px?

yorivanloo: True, solved.

PR #330, opened 2017-08-30
Add polyfill.io to mod_base <head>

rl-king:


modules/mod_ginger_base/templates/head/polyfill.tpl
```diff
@@ -0,0 +1,6 @@
+{#
+    More information can be found at https://polyfill.io/v2/docs/
+    Polyfill.io reads the User-Agent header of each request and returns polyfills that are suitable for the requesting browser.
+#}
+
+<script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
```
ddeboer: Do we need any of the polyfills for browsers other than IE11? If not, can we perhaps move this element to a conditional comment, thus only doing the HTTP requests for IE11 clients?

rl-king: I considered this. I noticed that if you include it like this, without queryparams, it will give you some polyfills when requesting from Safari 10.2.1 (a modern browser). If you always include this polyfill we won't have to worry about small modern browser support variations.

PR #338, opened 2017-10-04
[WIP] Added position property to map

yorivanloo:
[Base] Fixed scrollwheel setting value on true when opening infowindow

### DIFF
--- a/modules/mod_ginger_base/lib/js/map.js
+++ b/modules/mod_ginger_base/lib/js/map.js
@@ -8,6 +8,8 @@ $.widget( "ui.googlemap", {
 			id = widgetElement.attr('id'),
 			container = widgetElement,
 			locations = jQuery.parseJSON(widgetElement.data('locations')),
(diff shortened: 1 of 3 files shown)


modules/mod_ginger_base/lib/js/map.js
```diff
@@ -8,6 +8,8 @@ $.widget( "ui.googlemap", {
 			id = widgetElement.attr('id'),
 			container = widgetElement,
 			locations = jQuery.parseJSON(widgetElement.data('locations')),
+            infoBoxPosition = widgetElement.data('infoboxposition'),
+            scrollwheel = jQuery.parseJSON(widgetElement.data('scrollwheel')),
```
DorienD: Why do you use jQuery.parseJSON here?

yorivanloo: This line can be removed not needed, it was already in the map options data attribute

PR #341, opened 2017-10-12
Search suggestions update

rl-king:
- Added comments
- Re-ordered methods (which is why it looks like I rewrote _everything_)
- Break methods apart to improve readability
- Indicate which methods are public by `_` convention
- Removed `me = this` and use `this`
- Replaced `$.proxy` with `bind()`, which is native and recommended by jQuery as well
- Removed wrapping of widget in a closure, which had no effect
- Replaced `_init(){ this.init() };`, `_init(){ ... };`. jQueryUI calls `_created` and `_init`, I don't see the advantage of having a separate `init` over the default
- Move `base.js` eventlistener to `search-suggestions.js` since it was it's dependancy and makes more sense to be attached by itself
- Use element-ids declared at template `do_...` to get other elements from dom

### DIFF
--- a/modules/mod_ginger_base/lib/js/base.js
+++ b/modules/mod_ginger_base/lib/js/base.js
@@ -1,99 +1,60 @@
-(function($) {
-    'use strict';
+$.widget("ui.base", {
 
-    $.widget("ui.base", {
-
-        _init: function() {
-            this.init();
-        },
-
-        init: function() {
-            var me = this,
-                element = me.element;
-
-            //fancybox
-            $(".lightbox").fancybox({
-                'openEffect': "fade",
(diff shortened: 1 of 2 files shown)


modules/mod_ginger_base/lib/js/base.js
```diff
@@ -1,99 +1,60 @@
-(function($) {
```
ddeboer: Will removing this not cause any timing issues when JQuery hasn’t yet been fully loaded?

rl-king: Will check on this

rl-king:
A new version of this will be done in Elm

PR #347, opened 2017-11-09
New mod_ginger_embed

rl-king:
Replaces the Javascript generated embed code with an iframe. Sites are whitelisted in the site config with as comma separated list with an `allowed_hosts` key


DorienD:
@ddeboer This is now ready to be merged.

modules/mod_ginger_embed/dispatch/dispatch
```diff
@@ -1,4 +1,4 @@
 [
-    {embed_js, ["embed", "embed.js"], controller_file, [{path, "js/embed.js"}, {root, [lib]}]},
+    {embed, ["gingerembed", id], controller_template, [{template, "embed.tpl"}]},
```
ddeboer: Let’s not unnecessarily change the URL. The old one was fine (`["embed", id]`).

DorienD: agreed

DorienD: Done.

rl-king: Agreed, there might have been some reasoning about it at the time, not sure now. If this is reverted this https://github.com/driebit/ginger/blob/95fa7da2f2bd299745f82402b1341f560659209f/modules/mod_ginger_embed/mod_ginger_embed.erl#L66 has to be modified as well.

DorienD: shit! no we need ginger embed, because of the whitelisting

ddeboer: No, just change the above snippet to`/embedd`, too.

DorienD: ok

ddeboer:
Missing the primary reason for the changes in the description: CSS scoped to the individual embed. Adding this to the commit message.

ddeboer:
Thanks, all! We have a simpler embed now, unfortunately no longer used LD, but whatever.

PR #351, opened 2017-11-28
[base] Added filter for dynamic background position in masthead template

loetie:


DorienD:
Looks good!

modules/mod_ginger_base/filters/filter_background_position.erl
```diff
@@ -0,0 +1,58 @@
+%% @doc Force HTTPS URIs.
```
ddeboer: This seems wrong.

loetie:
@ddeboer it's fixed now

PR #357, opened 2018-01-29
[base] Fix search facet values state lost

ddeboer:
Keep facets' checked state (through the URL hash) between page loads.


ddeboer:
@robvandenbogaard This should fix our problem.

PR #358, opened 2018-01-30
[rdf] Improve resource to RDF

ddeboer:


modules/mod_ginger_rdf/models/m_rdf.erl
```diff
@@ -170,9 +170,12 @@ to_triples(Id, Context) ->
...
                                         object = m_rsc:p(Object, uri, Context)
                                     },
+                                    
+                                    %% Add literal triple with the edge's object
+                                    %% title for convenience.
                                     #triple{
-                                        predicate = PredicateUri,
-                                        subject = m_rsc:p(Subject, uri, Context),
+                                        predicate = <<?NS_DCTERMS, "title">>,
+                                        subject = m_rsc:p(Object, uri, Context),
```
ddeboer: @loetie This changes your code to return keywords. Can you check whether the keywords still function properly?

ddeboer: Tested and still works properly.

PR #360, opened 2018-02-08
[mod_ginger_rdf] RDF resources should not be owned by the user.

Dirklectisch:
I encountered some problems with RDF resources being owned and editable by users in one of our sites. I propose to have them alway be added by the site admin by default in an RDF content group. This way making a connection to an RDF resource will succeed no matter which user you are. The resources won't be editable unless you explicitly allow that in your site.


modules/mod_ginger_rdf/models/m_rdf.erl
```diff
@@ -103,14 +103,25 @@ ensure_resource(Uri, Props, Context) ->
 %% @doc Create non-authoritative RDF resource
 -spec create_resource(string(), list(), #context{}) -> integer().
 create_resource(Uri, Props, Context) ->
-    AllProps = [
+    AdminContext = z_acl:sudo(Context),
```
ddeboer: Can we do without this? I’d like to keep the original user as it provides some information about who originally created the RDF resource.

Dirklectisch: Why do we care about who made the resource initially? I think that's information that belongs on the edge not the resource in this case. The "creator" of a non authoritative resources is not the user but an external author.

DorienD:
What is the next step for this issue?

Dirklectisch:
I came to the conclusion that we should not give editing or insert rights to users for RDF resources. We do however want to limit who can do such a thing. Therefore I have now rewritten `m_rdf:ensure_resource/3` to check if the user is allowed to user the RDF module. Please give this another look @ddeboer .

** When this is deployed don't forget to give the appropriate rights in sites that use this module. **

modules/mod_ginger_rdf/models/m_rdf.erl
```diff
@@ -89,29 +89,51 @@ find_resource(Uri, Context) when is_integer(Uri) ->
...
+                    Props2 = z_utils:props_merge(Props1, DefaultProps),
+
+                    {ok, Id} = m_rsc_update:insert(Props2, AdminContext),
+                    Id;
+                Id ->
+                    {ok, Id} = m_rsc:update(Id, Props1, AdminContext),
+                    Id
+            end;
+        false ->
+            z_render:growl_error("Insufficient rights to update RDF resources", Context),
```
ddeboer: `?__()` to add this as a translation string.

modules/mod_ginger_rdf/models/m_rdf.erl
```diff
@@ -89,29 +89,51 @@ find_resource(Uri, Context) when is_integer(Uri) ->
...
+            % ... so that you can not insert non rdf resources this way.
+            RequiredProps = [
+                {category, rdf},
+                {is_authoritative, false},
+                {is_dependent, true}, %% remove resource when there are no longer edges to it
+                {uri, Uri}
+            ],
+            Props1 = z_utils:props_merge(RequiredProps, Props0),
+
+            AdminContext = z_acl:sudo(Context),
```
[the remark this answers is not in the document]
Dirklectisch: Yep, that's the idea.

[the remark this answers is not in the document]
ddeboer: Cool. Shouldn’t we add a deny ACL rule (or observer) here that actually does that? By default, editors are allowed to insert/update resources in all content groups.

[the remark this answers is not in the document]
Dirklectisch: As discussed in person, I added the block observer. I had to up the priority of the module to be the first observer in line.

modules/mod_ginger_rdf/models/m_rdf.erl
Dirklectisch: Fixed

modules/mod_ginger_rdf/models/m_rdf.erl
```diff
@@ -91,29 +91,51 @@ find_resource(Uri, Context) when is_integer(Uri) ->
...
+
+                    {ok, Id} = m_rsc_update:insert(Props2, AdminContext),
+                    Id;
+                Id ->
+                    {ok, Id} = m_rsc:update(Id, Props1, AdminContext),
+                    Id
+            end;
+        false ->
+            z_render:growl_error(?__("Insufficient rights to update RDF resources", Context), Context),
+            error
```
ddeboer: This doesn’t work: `z_render:growl_error/2` returns a new Context with the growl message added. By returning `error` after this, the returned Context is discarded, and therefore the growl message never shown.

A second reason why the growl message will not be shown is that `m_rdf_triple:insert/2` will now just crash anyway, because of:

```erlang
{ok, EdgeId} = m_rdf_triple:insert(Triple, Context),
```

Proposal: just return `{error, eacces}` here, which is Zotonic’s default error tuple for this case. Then let the caller (the event handlers in `m_ginger_rdf`) pattern match the return value of this function and add the growl message.

Dirklectisch:
@ddeboer I like to finish what I started, so here is another round of (hopefully the final) changes.

modules/mod_ginger_rdf/mod_ginger_rdf.erl
```diff
@@ -58,6 +64,22 @@ manage_schema(install, Context) ->
...
+        _ -> undefined
+    end;
+may_modify(_Id, _Context) ->
+    undefined.
+
+%% @doc Block editing of RDF resources bij default.
+observe_acl_is_allowed(#acl_is_allowed{action=insert, object=Id}, Context) ->
+    may_modify(Id, Context);
+observe_acl_is_allowed(#acl_is_allowed{action=update, object=Id}, Context) ->
+    may_modify(Id, Context);
```
ddeboer: These don’t need two cases.

ddeboer:
Thanks a lot! Your perseverance is appreciated.

PR #363, opened 2018-02-15
Improve masthead:

yorivanloo:
* More readable
* Stored recurring vars
* Add and include recurring element
* Move video check

Improve masthead:
* Some includes or checks where just to long so they came unreadable
* Also more readable, just better practise to store recurring vars
* 3 times the same element was implemented so made a seperate tpl, also when new content comes in elements the masthead.tpl does not become a mess
* Now the masthead.tpl was written that when a depiction vid was inserted in the cms this has 1st priority wich seems not most logic to me so i moved it to the depiction part and createn a check between image and video.


ddeboer:
@DorienD Do you think we should merge (parts of) this?

PR #364, opened 2018-02-19
[Base] Fix not working properly on iphone 5s (simulator)

yorivanloo:
The current check did not work in the xcode simulator i used and the new check works fine so maybe this one is more reliable?


DorienD:
@yorivanloo have you tested this on devices yet?

ddeboer:
Closing due to lack of feedback.

PR #361, opened 2018-02-08
[collection] ZZC-57 Added collection navigation

loetie:
Added collection navigation. Part of https://gitlab.driebit.nl/driebit/zuiderzeecollectie/merge_requests/10

PR #366, opened 2018-02-27
[collection] Fix missing pagelen argument in query offset

ddeboer:
* And don't add query_id to URL when query_id is null.

Ref #361.


modules/mod_ginger_collection/filters/filter_collection_query_offset.erl
```diff
@@ -9,8 +9,10 @@
 
 -include_lib("zotonic.hrl").
 
-collection_query_offset(Current, Pagelen, undefined, _Context) ->
-    get_offset(Current, Pagelen, 0);
+collection_query_offset(Current, Pagelen, undefined, Context) ->
+    collection_query_offset(Current, Pagelen, 1, Context);
+collection_query_offset(Current, undefined, Page, Context) ->
+    collection_query_offset(Current, 15, Page, Context);
```
ddeboer: @loetie Is it expected that `Pagelen` can sometimes be undefined? And if so, is `15` a good default?

modules/mod_ginger_collection/templates/collection/list/list-item.tpl
```diff
@@ -1,6 +1,6 @@
 {% with item._source as record %}
     <li class="list__item--beeldenzoeker {{ extraClasses }}">
-        <a href="{% block link %}{% url collection_object database=item._type object_id=record.priref %}?query_id={{ query_id }}&total={{ items.total }}&current={{ current|collection_query_offset:pagelen:page }}{% endblock %}"{% if link_target %} target="{{ link_target }}"{% endif %}>
+        <a href="{% block link %}{% url collection_object database=item._type object_id=record.priref %}{% if query_id %}?query_id={{ query_id }}&total={{ items.total }}&current={{ current|collection_query_offset:pagelen:page }}{% endif %}{% endblock %}"{% if link_target %} target="{{ link_target }}"{% endif %}>
```
ddeboer: This removes the `query_id=...` from the URL when if the collection object is not part of a query.

modules/mod_ginger_collection/filters/filter_collection_query_offset.erl
loetie: @ddeboer Set this as the default because in mod_ginger_collection on de page.beeldenzoeker_query.tpl this is used as the default pagelen. Not sure if it is actually needed.

PR #371, opened 2018-03-14
[base] Fix ginger_geo_nearby with negative lat/lng

### LINKED ISSUE #370: Negative lat or lng values breaks ginger_geo_nearby
For example with entoen.nu on acceptation server when using zipcode 1001AB.

This results into an order by like: 
`order by (pivot_location_lat-30.4226748)*(pivot_location_lat-30.4226748) + (pivot_location_lng--87.23802219999999)*(pivot_location_lng--87.23802219999999)`

and psql error message:

> syntax error at end of input

When using positive lat & lng the query succeeds.

ddeboer:
* Re-use Zotonic's geo search instead of duplicating the code.

Fix #370.

### DIFF
--- a/modules/mod_ginger_base/support/ginger_geo_search.erl
+++ b/modules/mod_ginger_base/support/ginger_geo_search.erl
@@ -19,45 +19,12 @@ search_query(#search_query{search={ginger_geo, Args}}, Context) ->
     };
 
 %% @doc Similar to z_geo_search:search_query/2 but result set includes locations and category
-search_query(#search_query{search={ginger_geo_nearby, Args}}, Context) ->
-
-    BaseArgs = lists:filter(
-        fun({Key, _}) ->
-            not lists:member(Key, [distance, lat, lng])
-        end,
-        Args
-    ),
-    BaseSearch = ginger_geo_search:search_query(#search_query{search={ginger_geo, BaseArgs}}, Context),
-
-    Distance = z_convert:to_float(proplists:get_value(distance, Args, 10)),
-    {Lat, Lng} = z_geo_search:get_query_center(Args, Context),
-    {LatMin, LngMin, LatMax, LngMax} = z_geo_support:get_lat_lng_bounds(Lat, Lng, Distance),
-
-    WhereStr = string:join(
-        [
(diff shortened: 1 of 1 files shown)


Dirklectisch:
@ddeboer I think your PR loses support for all `ginger_search` arguments doesn't it?

https://github.com/driebit/ginger/pull/371/commits/d372381e5459ec53218dc699a901d4303c6ff094#diff-40ce47d7d46c5f5b9332dab161106032L30

modules/mod_ginger_base/support/ginger_geo_search.erl
```diff
@@ -19,45 +19,49 @@ search_query(#search_query{search={ginger_geo, Args}}, Context) ->
...
-            z_convert:to_list(LngMax)
-        ],
-        " "
-    ),
-    SearchSql = and_where(BaseSearch, WhereStr),
+    %% Rebuild WHERE and ORDER BY because argument positions have shifted.
+    GeoWhere = arg(1, Placeholders) ++ " < pivot_location_lat AND " ++ arg(2, Placeholders) ++ "< pivot_location_lng "
+            ++ " AND pivot_location_lat < " ++ arg(3, Placeholders) ++ " AND pivot_location_lng < " ++ arg(4, Placeholders),
+    GeoOrder = "(pivot_location_lat-" ++ arg(5, Placeholders) ++ ")^2"
+        ++ " + (pivot_location_lng-" ++ arg(6, Placeholders) ++ ")^2",
```
ddeboer: This solves the actual bug: by keeping the coordinate floats as query parameters (instead of merging them into one query string), escaping any negative floats works properly.

ddeboer:
> I think your PR loses support for all ginger_search arguments doesn't it?

Ahum, yes, so it does. I think my drive to simplify turned into an oversimplification. 

Now adapted to keep the Ginger search params.

modules/mod_ginger_base/support/ginger_geo_search.erl
ddeboer: The original query has `A*A` but `A^2` is of course the same and doesn’t need the argument repetition.

PR #372, opened 2018-03-19
[base] Fix condition for showing rsvp button

robvandenbogaard:
The current condition for showing the RSVP button on an event's page leads to it only being visible on the last day of the event. This version seems to correct the logic: If the event is not finished yet, you can still add/change your RSVP.


ddeboer:
Thanks!

PR #380, opened 2018-04-19
[activity] Add activity inbox

ddeboer:
* Add HTTP API for getting and deleting items from the user's activity stream.
* BC break: #entry{} was renamed to #ginger_activity{}.
* Add documentation in readme.
* Apply mod_ginger_template template to README.md.

Depends on https://github.com/zotonic/zotonic/pull/1904 for DELETE requests.

### DIFF
--- a/modules/mod_ginger_activity/mod_ginger_activity.erl
+++ b/modules/mod_ginger_activity/mod_ginger_activity.erl
@@ -15,10 +15,17 @@
     init/1,
     event/2,
     register_activity/2,
+    register_activity/3,
     pid_observe_ginger_activity/3
 ]).
 
 -include("include/ginger_activity.hrl").
+
+-type activity() :: #ginger_activity{}.
+
+-export_type([
+    activity/0
+]).
 
 init(Context) ->
     case z_db:table_exists(activity_log, Context) of
@@ -59,7 +66,9 @@ init(Context) ->
                 create index activity_log_rsc_id_index on activity_log (rsc_id);
             ", Context),
             ok
-    end.
+    end,
+    m_ginger_activity:ensure_target_column(Context),
+    m_ginger_activity_inbox:init(Context).
 
 % @doc postback for activating resources
 event({postback,{activate, Args}, _TriggerId, _TargetId}, Context) ->
@@ -72,7 +81,12 @@ event({postback,{activate, Args}, _TriggerId, _TargetId}, Context) ->
     end.
 
 % @doc logical entry point for registering activity
-register_activity(RscId, Context) ->
+-spec register_activity(m_rsc:resource(), z:context()) -> ok.
+register_activity(Rsc, Context) ->
+    register_activity(Rsc, undefined, Context).
+
+-spec register_activity(m_rsc:resource(), m_rsc:resource(), z:context()) -> ok.
+register_activity(RscId, Target, Context) ->
     Time = calendar:universal_time(),
     UserId = z_acl:user(Context),
     IpAddress = case z_context:get_reqdata(Context) of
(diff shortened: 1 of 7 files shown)


modules/mod_ginger_activity/mod_ginger_activity.erl
```diff
@@ -85,28 +99,46 @@ register_activity(RscId, Context) ->
...
-    Props = [{rsc_id, RscId}, {time, DateTime}, {user_id, UserId}, {ip_address, IpAddress}],
-    z_db:insert(activity_log, Props, Context).
+-spec insert_activity(activity(), z:context()) -> ok.
+insert_activity(Activity, Context) ->
+    #ginger_activity{
+        rsc_id = Rsc,
+        target_id = Target,
+        time = Time,
+        user_id = User,
+        ip_address = Ip
```
robvandenbogaard: I know this was here before, but why does the ip address need to be logged?

mworrell:
@robvandenbogaard  IP addresses can be useful to remove activity when you identified a spammer. But that could also be done with a separate (more global) activity registration which can be truncated after a while.

modules/mod_ginger_activity/mod_ginger_activity.erl
```diff
@@ -85,28 +99,46 @@ register_activity(RscId, Context) ->
...
-% insert_activity(RscId, DateTime, UserId, Context) ->
-%     insert_activity(RscId, DateTime, UserId, undefined, Context).
-insert_activity(RscId, DateTime, UserId, IpAddress, Context) ->
-    Props = [{rsc_id, RscId}, {time, DateTime}, {user_id, UserId}, {ip_address, IpAddress}],
-    z_db:insert(activity_log, Props, Context).
+-spec insert_activity(activity(), z:context()) -> ok.
+insert_activity(Activity, Context) ->
+    #ginger_activity{
+        rsc_id = Rsc,
+        target_id = Target,
```
robvandenbogaard: target_id might be a confusing name, though I think I don't know a better alternative.
The target is an optional id of e.g. the page where resource Rsc was created. It provides a link to the context of Rsc which might be more useful to navigate to (instead of to the page for Rsc itself) when clicking the activity notification. So the "best" alternative I can think of would be "context_rsc_id". In its defense, the concept of "context of the activity" seems quite natural and useful.

modules/mod_ginger_activity/services/service_ginger_activity_inbox.erl
```diff
@@ -0,0 +1,84 @@
...
+    User = z_acl:user(Context),
+    case z_context:get_q(id, Context) of
+        undefined ->
+            %% Delete complete user's activity stream.
+            m_ginger_activity_inbox:delete(User, Context);
+        Id ->
+            IntId = z_convert:to_integer(Id),
+            m_ginger_activity_inbox:delete(User, IntId, Context)
+    end,
+    <<>>.
```
robvandenbogaard: I was wondering if we should return the data of the deleted item here, so ran a quick check on what is the common/best practice - there doesn't seem to be a hard one; https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#http-status for instance suggests returning http status 204 (success, no content) on delete without a body.

modules/mod_ginger_activity/services/service_ginger_activity_inbox.erl
```diff
@@ -0,0 +1,84 @@
...
+        <<"type">> => <<"Activity">>,
+        <<"published">> => Time,
+        <<"actor">> => #{
+            <<"type">> => <<"Person">>,
+            <<"name">> => z_trans:trans(m_rsc:p(Actor, title, Context), Context)
+        },
+        <<"object">> => #{
+            <<"id">> => m_rsc:p(Object, page_url_abs, Context),
+            <<"type">> => m_rsc:p(m_rsc:p(Object, category_id, Context), name, Context),
+            <<"name">> => z_trans:trans(m_rsc:p(Object, title, Context), Context),
```
robvandenbogaard: I am not sure I will have a use for the Object's title in my use case (only showing the content crafted at activity log creation for the Object), but in general it might be useful.

PR #381, opened 2018-04-20
AVG reminder in admin

DorienD:
User gets a banner on the bottom with the reminder to check their privacy policy. They can close this reminder and a cookie will remember this.


loetie:
@DorienD Title of pull request still says "WIP". Is that correct or not? Changes seem fine to me.

Issue #167, opened 2016-09-01
Support for multilingual RDF content


ddeboer:
See the [JSON-LD spec](https://www.w3.org/TR/json-ld/#string-internationalization).

Currently, mod_ginger_rdf only returns the content in the default language (as long as the  `mod_translation.force_default` config key is set).

ddeboer:
RDF export was made multilingual in #404.

PR #382, opened 2018-04-23
[rdf] Add tests to prepare for literal object value translations fix

robvandenbogaard:


robvandenbogaard:
This might fix #167 if it handles language settings accordingly.

PR #383, opened 2018-04-23
[rdf] Support for translated literal object values

robvandenbogaard:
This is to support RDF literal object values with multiple translations, which are currently just appended in a list. This pull request adds selection of the translations matching the current or default language.

### DIFF
--- a/modules/mod_ginger_rdf/tests/m_rdf_tests.erl
+++ b/modules/mod_ginger_rdf/tests/m_rdf_tests.erl
@@ -1,0 +1,98 @@
+-module(m_rdf_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+-include("zotonic.hrl").
+-include("../include/rdf.hrl").
+
+m_find_value_test() ->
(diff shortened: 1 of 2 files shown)


modules/mod_ginger_rdf/tests/m_rdf_tests.erl
```diff
@@ -0,0 +1,98 @@
...
+                subject = <<"http://dinges.com/123">>,
+                predicate = <<"dc:author">>,
+                object = [
+                    #rdf_value{
+                        value = <<"Author 1">>
+                    },
+                    #rdf_value{
+                        value = <<"Author 2">>
+                    }
+                ]
```
ddeboer: This is not correct: currently a triple’s object is defined as

```erlang
object :: binary() | #rdf_value{},
```

so either a single URI (binary) or a single `#rdf_value{}`, but never a list.

DorienD:
@robvandenbogaard @ddeboer is this still relevant?

PR #393, opened 2018-06-06
[activity] Publish MQTT message when activity is added to user's stream

ddeboer:


modules/mod_ginger_activity/models/m_ginger_activity_inbox.erl
```diff
@@ -86,7 +86,9 @@ insert(ActivityId, UserId, Context) ->
             {user_id, UserId}
         ],
         Context
-    ).
+    ),
+    Topic = <<"~site/user/", (z_convert:to_binary(UserId))/binary, "/activities">>,
```
ddeboer: @robvandenbogaard I had to change the topic from `~session` to `~user`, which is actually more correct.

PR #394, opened 2018-06-12
Rest API

ddeboer:


docs/rest-api.raml
```diff
@@ -0,0 +1,50 @@
...
+                  schema:
+                      type: integer
+                  required: true
+    /resources/path/{path}:
+        get:
+            parameters:
+                - in: path
+                  schema:
+                      type: integer
+                  required: true
```
ddeboer: How do these definitions compare to JSON-LD, the format that we want to use for our API? 

> As a result, different sites have chosen their own proprietary representations for interacting with their sites, sometimes described using frameworks such as [swagger] which imply a particular URI composition for interacting with their services. This practice leads to vendor-specific semantic silos, where the meaning of a particular JSON document makes sense only by programming directly to the API documentation for a given service.

https://json-ld.org/spec/latest/json-ld-api-best-practices/

docs/rest-api.yaml
```diff
@@ -0,0 +1,106 @@
+swagger: "2.0"
```
ddeboer: Do we need both raml and yaml?

modules/mod_ginger_base/tests/mod_ginger_base_rest_tests.erl
```diff
@@ -0,0 +1,20 @@
...
+             State = #state{mode = collection},
+             {Result, _, _} =  controller_rest:resource_exists(req, State),
+             ?assertEqual(true, Result)
+     end,
+     fun () ->
+             State = #state{mode = document},
+             {Result, _, _} =  controller_rest:resource_exists(req, State),
+             ?assertEqual(false, Result)
+     end
+    ].
```
ddeboer: Let’s also add integration tests against the sandbox here.

modules/mod_ginger_base/tests/mod_ginger_base_rest_tests.erl
```diff
@@ -0,0 +1,20 @@
+-module(mod_ginger_base_rest_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+-include("zotonic.hrl").
+
+-record(state, {mode}).
+
+resource_exists_test_() ->
+    [
+     fun () ->
```
ddeboer: No space between `fun` and parens. While not explicitly mentioned by Inaka’s guidelines, let’s stick to [their practice](https://github.com/inaka/erlang_guidelines/blob/e2b67c0d48c7348185e6584f9636c12a777277ef/src/import.erl#L9).

rmies: Hmm, that doesn't make sense to me; the form is `fun [Name]...` (see http://erlang.org/doc/reference_manual/expressions.html#id84308) where `Name` is optional, therefore I prefer to do have a space between `fun` and `()`.

docs/rest-api.yaml
rmies: No, we switched from RAML to [OpenAPI](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md).

modules/mod_ginger_base/tests/mod_ginger_base_rest_tests.erl
rmies: Also note that the Erlang reference manual *does* a space.

docs/rest-api.yaml
rl-king: https://github.com/driebit/ginger/pull/394/commits/0daf23fe77324866321b09605fd7b6fcec4f420c

modules/mod_ginger_base/support/types/ginger_boolean.erl
```diff
@@ -0,0 +1,11 @@
...
+
+-export([encode/1]).
+
+-include("zotonic.hrl").
+
+-spec encode(term()) -> boolean().
+encode(Value) when is_boolean(Value) ->
+    Value;
+encode(Value) ->
+    erlang:error(<<"Expecting a true or false atom">>, Value).
```
ddeboer: We decided to crash, so no need to add the fallback case here.

modules/mod_ginger_base/tests/mod_ginger_base_rest_tests.erl
ddeboer: Yep, but we follow Inaka’s guidelines.

PR #401, opened 2018-06-21
[rdf] Add functions for constructing RDF properties

ddeboer:
* Prefer function calls over macros.

### DIFF
--- a/modules/mod_ginger_rdf/support/rdf_property.erl
+++ b/modules/mod_ginger_rdf/support/rdf_property.erl
@@ -4,20 +4,44 @@
 -export([
     'dbpedia-owl'/1,
     foaf/1,
-    rdfs/1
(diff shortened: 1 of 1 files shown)


modules/mod_ginger_rdf/support/rdf_property.erl
```diff
@@ -4,20 +4,44 @@
...
+    rdf/1,
+    rdfs/1,
+    schema/1,
+    vcard/1
 ]).
 
 -include_lib("zotonic.hrl").
 -include_lib("../include/rdf.hrl").
 
+-spec 'dbpedia-owl'(binary()) -> binary().
```
rmies: Is it possible to be more "expressive" here (and below) by using a type that is an alias for `binary()`?

PR #404, opened 2018-06-26
[rdf] Add schema.org RDF serialization

ddeboer:
* Move RDF export logic into new m_rdf_export module.
* Add rdf_ontology behaviour and schema_org as the first implementation.
* Export content in all available languages instead of only the default one.
* Add more tests.

### DIFF
--- a/modules/mod_ginger_rdf/models/m_rdf.erl
+++ b/modules/mod_ginger_rdf/models/m_rdf.erl
@@ -20,14 +20,18 @@
     ensure_resource/3,
     ensure_resource_edges/4,
     lookup_triple/2,
-    to_triples/2
+    to_triples/2,
+    rdf_resource/2
 ]).
 
 -opaque rdf_resource() :: #rdf_resource{}.
 -opaque triple() :: #triple{}.
+-type predicate() :: ginger_uri:uri().
+
 -export_type([
     rdf_resource/0,
-    triple/0
+    triple/0,
+    predicate/0
 ]).
 
 m_find_value(#rdf_resource{} = Rdf, #m{value = undefined} = M, _Context) ->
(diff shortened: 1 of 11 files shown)


modules/mod_ginger_rdf/templates/mediaclass.config
```diff
@@ -1,5 +1,5 @@
 [
-    {"foaf-thumbnail", [
+    {"rdf-thumbnail", [
```
ddeboer: Renamed this as it’s not about FOAF specifically, but RDF generally.

modules/mod_ginger_rdf/models/m_rdf_export.erl
```diff
@@ -0,0 +1,161 @@
...
+to_rdf(Id, Context) ->
+    to_rdf(Id, [schema_org], Context).
+
+%% @doc Export a Zotonic resource to RDF.
+%%      Combine:
+%%      - properties
+%%      - outgoing edges
+%%      - media.
+-spec to_rdf(m_rsc:resource(), [module()], z:context()) -> m_rdf:rdf_resource().
+to_rdf(Id, Ontologies, Context) ->
```
ddeboer: Pass in a list of ontologies for dependency injection.

modules/mod_ginger_rdf/models/m_rdf_export.erl
```diff
@@ -0,0 +1,161 @@
...
+    with_original_media/4,
+    with_thumbnail/4
+]).
+
+-include_lib("zotonic.hrl").
+-include_lib("../include/rdf.hrl").
+
+-spec to_rdf(m_rsc:resource(), z:context()) -> m_rdf:rdf_resource().
+to_rdf(Id, Context) ->
+    to_rdf(Id, [schema_org], Context).
```
ddeboer: `schema_org` is a sane default, as it’s what search engines like to have.

modules/mod_ginger_rdf/models/m_rdf.erl
```diff
@@ -252,160 +191,6 @@ rdf_resource(SubjectUri, Triples) ->
...
-property_to_triples({address_postcode, Value}, _Props, _Context) ->
-    [
-        #triple{predicate = <<?NS_VCARD, "postal-code">>, object = #rdf_value{value = Value}}
-    ];
-property_to_triples({address_street_1, Value}, _Props, _Context) ->
-    [
-        #triple{predicate = <<?NS_VCARD, "street-address">>, object = #rdf_value{value = Value}}
-    ];
-property_to_triples({body, Value}, _Props, Context) ->
-    %% TODO add support for multilingual properties
```
ddeboer: This was resolved.

PR #417, opened 2018-07-04
Convert path ids to integers

rl-king:
Patternmatching the `id` from the path will return a string, we want an integer.

### DIFF
--- a/modules/mod_ginger_base/controllers/controller_rest.erl
+++ b/modules/mod_ginger_base/controllers/controller_rest.erl
@@ -202,9 +202,9 @@ path_to_id("/", Context) ->
 path_to_id(Path, Context) ->
     case string:tokens(Path, "/") of
         ["page", Id | _] ->
-            {ok, Id};
+            {ok, erlang:list_to_integer(Id)};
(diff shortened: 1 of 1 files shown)


modules/mod_ginger_base/controllers/controller_rest.erl
```diff
@@ -202,9 +202,9 @@ path_to_id("/", Context) ->
 path_to_id(Path, Context) ->
     case string:tokens(Path, "/") of
         ["page", Id | _] ->
-            {ok, Id};
+            {ok, erlang:list_to_integer(Id)};
         [_Language, "page", Id | _] ->
-            {ok, Id};
+            {ok, erlang:list_to_integer(Id)};
```
ddeboer: Why process this manually and not just always call `m_rsc:page_path_to_id/2`, as in the case below?

rl-king: I believe paths like `/page/348/mypage` are not stored in the DB and `m_rsc:page_path_to_id/2` is a DB lookup, so it only works if it's deliberately set.

PR #418, opened 2018-07-05
[rdf] Improve SPARQL queries for use on DBpedia

ddeboer:


modules/mod_ginger_rdf/support/sparql_client.erl
```diff
@@ -68,4 +95,4 @@ decode(Data) ->
     Data.
 
 headers() ->
-    #{<<"Accept">> => <<"application/ld+json">>}.
+    #{<<"Accept">> => <<"application/json">>}.
```
robvandenbogaard: Should we make this: `<<"application/ld+json,application/json;q=0.9">>`?

modules/mod_ginger_rdf/support/sparql_query.erl
```diff
@@ -2,48 +2,100 @@
...
+%% @doc Add a query argument.
+-spec add_argument(binary(), sparql_query()) -> sparql_query().
+add_argument(Property, #sparql_query{} = Query) ->
+    #sparql_query{select = Select, arguments = Arguments, where = Where} = Query,
+    Next = z_convert:to_binary(maps:size(Arguments) + 1),
+    Argument = <<"?", Next/binary>>,
+    Query#sparql_query{
+        select = Select ++ [Argument],
+        arguments = Arguments#{Next => Property},
+        where = [<<"OPTIONAL {?s <", Property/binary, "> ?", Next/binary, "}">> | Where]
```
robvandenbogaard: (Aside) Am I correct to conclude that predicates are expected to always be in url form, hence the angular braces? E.g. when using schema prefixes the braces should not be there, as in `schema:name` or `a` (built-in predicate, maybe specific to Virtuoso).

robvandenbogaard: Tests are unchanged, we'd probably like to have some extra for the query builder.

modules/mod_ginger_rdf/support/sparql_client.erl
```diff
@@ -42,20 +46,43 @@ query(Endpoint, Query, Headers) ->
...
+    query_rdf(Endpoint, Query, headers()).
+
+%% @doc Execute query and return result as RDF.
+-spec query_rdf(ginger_uri:uri(), query(), map()) -> [m_rdf:rdf_resource()].
+query_rdf(Endpoint, Query, Headers) ->
+    case query(Endpoint, Query, Headers) of
+        undefined ->
+            undefined;
+        #{<<"results">> := #{<<"bindings">> := Bindings}} ->
+            %% Result set can have multiple rows with same ?s (subject URI), so combine those into
```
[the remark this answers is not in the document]
ddeboer: The problem is that pagination usually assumes that you’re iterating over distinct subjects. Exactly in the case where the SPARQL query returns duplicate rows, pagination seems off. And remember that we want to return a list of `#rdf_resource{}`s here, which should by definition be about a single subject (`#rdf_resource.id`).

ddeboer:
> I am wary of resource merging of results, but I might overlook that this handles cases where it is not a problem.

The merge is never destructive: all triples are preserved, only exact duplicates are removed.

modules/mod_ginger_rdf/support/sparql_client.erl
ddeboer: Yes, better. I’ll have to test that header against the DBpedia endpoint first.

modules/mod_ginger_rdf/support/sparql_query.erl
ddeboer: > Am I correct to conclude that predicates are expected to always be in url form

For now, yes. An improvement would to pattern match on `http(s)://` and only then add the guillemets `< >`.

modules/mod_ginger_rdf/support/sparql_client.erl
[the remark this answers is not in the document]
ddeboer: Removed results merging.

modules/mod_ginger_rdf/support/sparql_client.erl
ddeboer: Unfortunately Virtuoso doesn’t get Accept headers properly, so had to work around this.

ddeboer:
Adapted to your comments @robvandenbogaard.

PR #421, opened 2018-07-18
[rest] Exclude empty dates from REST output

ddeboer:


modules/mod_ginger_base/controllers/controller_rest.erl
```diff
@@ -162,12 +162,13 @@ custom_props(Id, Context) ->
...
             maps:fold(
               fun(PropName, TypeModule, Acc) ->
-                      case m_rsc:p(Id, PropName, Context) of
-                          undefined ->
-                              Acc;
-                          Value ->
-                              Acc#{PropName => TypeModule:encode(Value)}
-                      end
+                  Value = m_rsc:p(Id, PropName, Context),
+                  case z_utils:is_empty(Value) of
```
ddeboer: This also checks for `?ST_JUTTEMIS`.

PR #433, opened 2018-09-03
Bump Elasticsearch version in Docker Compose file

linuss:
The previous elasticsearch version crashes almost immediatly on my system.
I don't know the exact cause, but a (small) version bump fixed the issue.
Because we use ES v5.3.2 in our production environments too I picked that
version to use here.


ddeboer:
Cool, thanks!

PR #435, opened 2018-09-05
[base] Add cat_promote search argument for boosting categories

ddeboer:


rmies:
Is there documentation about the search options where we need to make an addition for this extension?

ddeboer:
Nope.

rmies:
Okay, then I suggest we add documentation.

ddeboer:
#436

PR #438, opened 2018-09-10
Add make target to create API doc

rmies:


Makefile
```diff
@@ -30,6 +33,10 @@ help:
 addsite:
 	@docker-compose exec zotonic bin/zotonic addsite -s ginger -H $(name).docker.test $(name)
 
+api-docs:
+	@yaml2json docs/rest-api.yaml > tmp/rest-api.json
```
ddeboer: Why not simply a local `tmp` dir, such as `/tmp`? It seems unneeded to check in `tmp/` into the repository.

Makefile
```diff
@@ -8,6 +8,9 @@ target test-chrome: url=http://$(site).docker.test
 
 include .env
 
+NPM_PATH := ./node_modules/.bin
+export PATH := $(NPM_PATH):$(PATH)
+
```
ddeboer: A better way is to add the `api-docs` command as a `script` in `package.json`. That way, these lines here can be removed.

rmies: I don't see how this is better. We already use the above construction in several other places, so for the sake of consistency I would like to keep it this way.

Makefile
```diff
@@ -30,6 +33,10 @@ help:
 addsite:
 	@docker-compose exec zotonic bin/zotonic addsite -s ginger -H $(name).docker.test $(name)
 
+api-docs:
+	@yaml2json docs/rest-api.yaml > /tmp/rest-api.json
+	@spectacle -1 -t /tmp -f ginger-rest-api.html /tmp/rest-api.json
```
ddeboer: Should we add a `deps` step to Ginger’s Makefile, too? Otherwise `make api-docs` doesn’t work out of the box, as it needs to be preceded by an `npm install`.

PR #441, opened 2018-09-11
[base] Return empty list of properties as null instead of {}

ddeboer:


ddeboer:
The `{}` was hurting my eyes.

PR #447, opened 2018-09-13
[base] Add range facet to search API

ddeboer:


modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -102,4 +104,37 @@ whitelisted(Arguments) ->
...
+    maps:fold(
+        fun(K, V, Acc) ->
+            Skip = byte_size(K) - 4,
+            Seven = byte_size(K) - 7,
+            case K of
+                <<Field:Skip/binary, "_min">> ->
+                    Acc#{Field => maps:put(<<"min">>, V, maps:get(Field, Acc, #{}))};
+                <<Field:Skip/binary, "_max">> ->
+                    Acc#{Field => maps:put(<<"max">>, V, maps:get(Field, Acc, #{}))};
+                <<Field:Seven/binary, "_global">> ->
```
ddeboer: @rmies These patterns are here to match `_min|_max` and `_global`, reactive to the output received from `z_search:search/3`. These patterns look somewhat messy. An alternative is to use the user-supplied list of facets (as query string param) as a starting point. Do you have any suggestions to clean this up?

rmies: I'm afraid not.

ddeboer:
Resolved merge conflicts.

PR #450, opened 2018-09-25
[base] Create coordinates endpoint

linuss:


modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -2,10 +2,10 @@
...
 
 -export([
-    init/1,
-    content_types_provided/2,
-    to_json/2
-]).
+	 init/1,
+	 content_types_provided/2,
+	 to_json/2
+	]).
```
ddeboer: Please don’t make unnecessary whitespace changes. The old code matches [Inaka’s guidelines](https://github.com/inaka/erlang_guidelines), which we follow.

modules/mod_ginger_base/models/m_ginger_rest.erl
```diff
@@ -64,11 +64,14 @@ with_edges(Rsc = #{<<"id">> := Id}, Predicates, Context) ->
...
+-spec translations(binary() | {trans, proplists:proplist()}, z:context()) -> translations().
 translations({trans, Translations}, Context) ->
-    [{Key, z_html:unescape(filter_show_media:show_media(Value, Context))} || {Key, Value} <- Translations];
+    lists:foldl(fun({Key, Value}, Acc) -> 
+		       V = z_html:unescape(filter_show_media:show_media(Value, Context)),
+		       Acc#{Key => V}
+	       end, #{}, Translations);
 translations(Value, Context) ->
-    [{z_trans:default_language(Context), Value}].
+    #{z_trans:default_language(Context) => Value}.
```
ddeboer: 👍 but this looks unrelated to the scope of this PR. After all, in the coordinates search result no translations are returned (and therefore encoded/decoded) at all.

modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -16,36 +16,89 @@ init(Args) ->
...
+	    Query2 = [{has_geo, <<"true">>} | Query1],
+	    SearchResults = z_search:search({Type, Query2}, {Offset + 1, Limit}, Context),
+            %% Strip any unneeded data
+	    Coordinates = 
+                lists:map(
+                  fun(Item) -> 
+                          Id = maps:get(<<"_id">>, Item),
+                          #{<<"lat">> := Lat,
+                            <<"lon">> := Lon} = 
+                              get_in([<<"_source">>, <<"geolocation">>], Item),
```
ddeboer: This works, but what’s wrong with a pattern match in `fun(Item)` instead? To me, it would be clearer.

modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -16,36 +16,89 @@ init(Args) ->
...
+    lists:map(fun(V) -> json_map(F, V) end, Elm);
+json_map(F, Elm) ->
+    F(Elm).
+
+%% @doc Converts a Erlang formatted date-time to an ISO-8601 string. 
+%% Simply returns anything else
+stringify_date({{_Y, _M, _D},{_H, _Mi, _S}} = Date, Context) ->
+    z_datetime:format(Date, "c", Context);
+stringify_date(V, _Context) ->
+    V.
```
ddeboer: Okay for now, but mixing concerns. Should be moved to a dedicated JSON module in the future.

modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -20,32 +20,82 @@ to_json(Req, State) ->
...
-    Json = jsx:encode(SearchResults),
-    %% Done
-    {Json, Req, State}.
+    case proplists:get_value(mode, State) of 
+        coordinates ->
+            %% We're only interested in the geolocation and id fields
+	    Query1 = [{source, [<<"geolocation">>]} | arguments(RequestArgs)],
+            %% And it doesn't make sense to retrieve any results without coordinates
+	    Query2 = [{has_geo, <<"true">>} | Query1],
+	    SearchResults = z_search:search({Type, Query2}, {Offset + 1, Limit}, Context),
```
ddeboer: Maybe move the comment above (`Zotonic offsets start at 1`) here, because it explains the `Offset +1`.

modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -20,32 +20,63 @@ to_json(Req, State) ->
...
-        <<"total">> => Total
-    },
-    Json = jsx:encode(SearchResults),
-    %% Done
-    {Json, Req, State}.
+    
+    case proplists:get_value(mode, State) of 
+        coordinates ->
+            %% We're only interested in the geolocation and id fields
+	    Query1 = [{source, [<<"geolocation">>]} | arguments(RequestArgs)],
```
ddeboer: Indent looks really off here: how comes this aligns to the _left_ of `coordinates ->` above?

PR #457, opened 2018-10-04
[embed] Fix duplicate key html_elt_extra

### LINKED ISSUE #453: mod_ginger_remark forces site language to `nl`
To me it is unexpected that having `mod_ginger_remark` enabled will force the site's language to `nl`.

https://github.com/driebit/ginger/blob/master/modules/mod_ginger_remark/mod_ginger_remark.erl#L26

```erlang
manage_schema(_Version, Context) ->
    m_config:set_value(i18n, language, nl, Context),
    Datamodel = #datamodel{
```

Maybe that line should be removed?

ddeboer:
This fixes: duplicate key value violates unique constraint
\"config_module_key_key\"">>,[{detail,
<<"Key (module, key)=(site, html_elt_extra) already exists.">>}]

This is already configured in mod_ginger_base, so no need to duplicate
it here.

### DIFF
--- a/modules/mod_ginger_embed/mod_ginger_embed.erl
+++ b/modules/mod_ginger_embed/mod_ginger_embed.erl
@@ -21,8 +21,6 @@
 -include("zotonic.hrl").
 
 manage_schema(_Version, Context) ->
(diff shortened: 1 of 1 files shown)


ddeboer:
Ping @DorienD.

PR #460, opened 2018-10-05
[base] Add controller to allow logins from SPAs (e.g. Elm)

rmies:
Work in progress and when it is finished we need to discuss security implications.


rl-king:
We probably want to have a mechanism/endpoint which returns the current authentication status as well

PR #462, opened 2018-10-18
[edit] remove option to redirect on page view on person page

DorienD:


modules/mod_ginger_edit/templates/_ginger_edit_links.tpl
```diff
@@ -30,15 +30,17 @@
...
-            <div class="form-group checkbox">
-                <label>
-                    <input type="checkbox" id="field-is-website=redirect" name="is_website_redirect" value="1"
-                        {% if r.is_website_redirect %}checked{% endif %}
-                        {% if not is_editable %}disabled="disabled"{% endif %}
-                    />
-                    {_ Redirect to website on page view _}
-                </label>
-            </div>
+            {% if not id.category_id == m.rsc.person.id %}
```
ddeboer: Why not:

```suggestion
            {% if not id|is_a:"person" %}
```

DorienD: If we allow this, then users can redirect their own profile page to another website and will not be able to edit their profile page. Since they are redirected to that other page.

PR #470, opened 2018-11-01
[base] Disable the noindex checkbox in the admin

ddeboer:
* Read the noindex param only from the site/config file, not from the 
  database.
* When disabled, saving the form will not persist a value for noindex
  in the database. This means the value from site/config will always be
  leading.

### DIFF
--- a/modules/mod_ginger_base/templates/admin_seo.tpl
+++ b/modules/mod_ginger_base/templates/admin_seo.tpl
@@ -1,0 +1,109 @@
+{% overrules %}
+
+{% block content %}
+
+    <div class="admin-header">
+        <h2>{_ Search Engine Optimization _}</h2>
+        <p>{_ Here you find settings to optimize the indexing of your site by search engines. _}</p>
+    </div>
+
+    {% wire id="admin-seo" type="submit" postback="admin_seo" %}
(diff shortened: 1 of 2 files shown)


modules/mod_ginger_base/templates/admin_seo.tpl
```diff
@@ -0,0 +1,109 @@
...
+                                    {_ Description to include on every page _}
+                                </p>
+                            </div>
+                        </div>
+
+                        <div class="form-group row">
+                            <div class="col-md-8 col-md-offset-4">
+                                <label class="checkbox-inline" for="seo-noindex" title="{_ Add a noindex, nofollow element to all pages. _}">
+                                    <input type="checkbox" id="seo-noindex" name="seo-noindex" value="1" {% if m.config.seo.noindex.value %}checked="checked"{% endif %} disabled />
+                                    {_ Exclude this site from search engines _}. {_ This setting can only be changed in the site config file. _}
```
ddeboer: Opted to show a disabled checkbox instead of removing it altogether, so people won’t be looking around for something they can’t find. 😉

CyBeRoni: I prefer that to removing it, too.

ddeboer:
So merge?

PR #481, opened 2018-11-15
Run all tests on CI

ddeboer:

PR #489, opened 2018-11-26
[base] Allow deletion of edges via REST API

rmies:


ddeboer:
Do these tests run on CI? Cf. #481.

rmies:
Why ask if you already know the answer?

PR #496, opened 2018-12-04
Create module for uploading resources to SWORD endpoints

linuss:


modules/mod_ginger_sword/mod_ginger_sword.erl
```diff
@@ -0,0 +1,232 @@
...
+format_rsc_uri(RscId, Context) ->
+    sword_url(Context) ++ "/" ++ format_slug(RscId, Context).
+
+render_rdf_body(RscId, Context) ->
+    AC = z_acl:sudo(Context), % We want to publish all properties even if hidden
+    case m_rsc:exists(RscId, Context) of
+        true ->
+            RdfResource = m_rdf:to_triples(RscId, AC),
+            JsonLd = ginger_json_ld:serialize(RdfResource),
+            iolist_to_binary(mochijson2:encode(JsonLd));
```
ddeboer: Why not use `jsx:encode/1`?

modules/mod_ginger_sword/mod_ginger_sword.erl
```diff
@@ -0,0 +1,232 @@
...
+    {Status, _} = Resp,
+    case Status of
+        ok ->
+            Resp;
+        _ ->
+            lager:warning("Http request failed"),
+            lager:info("~p~n", [Request]),
+            lager:info("~p~n", [Resp]),
+            Resp
+    end.
```
ddeboer: You can use `ginger_http_client` instead for better separation of concerns.

modules/mod_ginger_sword/mod_ginger_sword.erl
```diff
@@ -0,0 +1,232 @@
...
+                            Context).
+
+%% @doc Takes a category path and determines if it must be published
+cat_must_publish(IsA) ->
+    ExactCat = case IsA of
+                   [] ->
+                       undefined;
+                   _ ->
+                       lists:last(IsA)
+               end,
```
ddeboer: I would replace this `case` with a pattern match: `cat_must_publish([])`. See also https://github.com/inaka/erlang_guidelines#avoid-unnecesary-calls-to-length1 (although you’re not using `length/1`, the example is similar).

modules/mod_ginger_sword/mod_ginger_sword.erl
```diff
@@ -0,0 +1,232 @@
...
+    lists:any(
+      fun(Cat) ->
+              ExactCat == Cat
+      end,
+      [story]
+     ).
+
+%% @doc Determines if a resource state is published (to the aggregator)
```
ddeboer: Rather s/is/must be published, right?

modules/mod_ginger_sword/mod_ginger_sword.erl
linuss: Good one, changed!

modules/mod_ginger_sword/mod_ginger_sword.erl
linuss: Done!

modules/mod_ginger_sword/mod_ginger_sword.erl
linuss: Check out the update, it's totally gone now :)

modules/mod_ginger_sword/mod_ginger_sword.erl
linuss: Method removed, no longer applicable :)

PR #516, opened 2019-01-14
Change password reset form to be compatible with ratelimit checks in 0.x

mworrell:
Authentication postbacks in 0.x have been changed to accommodate ratelimit checks on logon and password reset actions.


mworrell:
The Zotonic template is the same as the Ginger template, so I changed this MR to remove the Ginger version.

ddeboer:
Thanks!

PR #518, opened 2019-01-15
Stop error when global nav does not require suggestions

DorienD:


ddeboer:
What error were you getting?

PR #523, opened 2019-01-23
[doc] Document selective Docker

ddeboer:
This adds a hybrid Docker setup that is compatible with the full Docker setup. Zotonic itself runs directly on the host, communicating with containers running Postgres and Elasticsearch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,10 +88,10 @@ source code, too.
 All other services (PostgreSQL, Elasticsearch, Kibana) still run in
 containers.
 
-First, install Install Erlang locally:
+First, install install Erlang locally:
 
 ```bash
-$ brew install erlang@20
+$ brew install erlang@20 imagemagick
 $ brew link erlang@20 --force
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -66,6 +75,58 @@ $ make && bin/zotonic runtests mod_ginger_collection
...
+
+```bash
+$ git clone https://github.com/driebit/ginger.git
+$ git clone https://github.com/zotonic/zotonic.git --branch 0.x 
+```
+
+Copy Ginger’s configuration file, which includes its dependencies:
+
+```bash
+cp ginger/config/zotonic.config ~/.zotonic/0/zotonic.config
```
rl-king: Create dir `~/.zotonic/0/` if missing? Assuming it is, since you've never started Zotonic

README.md
```diff
@@ -66,6 +75,58 @@ $ make && bin/zotonic runtests mod_ginger_collection
...
+And run Zotonic:
+
+```bash
+$ cd ginger
+$ make up-support
+$ cd ../zotonic
+$ bin/zotonic debug
+```
+
+Zotonic then runs on port 8000: http://localhost:8000.
```
rl-king: Skel uses to site.docker.test https://github.com/driebit/ginger/blob/master/skel/config.in#L3

Add note on how to solve this? 
```echo "rdr pass inet proto tcp from any to any port 80 -> 127.0.0.1 port 8000" | sudo pfctl -ef -```

ddeboer: Resolved with a `make start` target that both runs `pfctl` and starts/stops Docker containers.

ddeboer:
> I remember having to compile jsx with export JSX_FORCE_MAPS=1, might be applicable here

Good point! Added.

ddeboer:
Thanks for reviewing!

PR #545, opened 2019-02-22
[base] Remove header and style from blocks

rl-king:
There is no method of filling these fields so no reason include these in the response


ddeboer:
@rl-king You should also update the (now failing) tests: https://github.com/driebit/ginger/blob/4bf32af36c8539c23ca3ce48a33f99b9e4cfcd70/modules/mod_ginger_base/tests/m_ginger_rest_tests.erl#L37

rl-king:
Done

PR #546, opened 2019-02-24
[Remark] remove escape, caused error in getting properties of remark

DorienD:


modules/mod_ginger_remark/templates/remark/remark.tpl
```diff
@@ -1,7 +1,6 @@
-
 {% with
     editing|default:q.editing|default:0|escape,
-    remark_id|default:q.remark_id|escape,
+    remark_id|default:q.remark_id,
```
ddeboer: This escape was added in #530 for security reasons, so we shouldn’t just remove it.

@DorienD What error are you getting exactly?

DorienD: The remarks are not showing any content: http://hartamsterdam.ginger-acceptatie.driebit.net/ and also on https://geheugenvanoost.ginger-acceptatie.driebit.net/page/57026/the-band. When i removed this escape locally it worked. 

I've now added a different parameter to the include, this also seems to work. So i've reverted the removal of the escape.

ddeboer: That’s better!

PR #549, opened 2019-02-26
Add correct caption fallbacks for audio

DorienD:


modules/mod_ginger_base/templates/media/media.audio.tpl
```diff
@@ -5,19 +5,15 @@ as
...
 
-    <div>
-        {% if id.summary %}
-            {{ id.summary }}
-        {% elif id.title %}
-            {{ id.title }}
-        {% endif %}
-    </div>
-
+    {% if caption|default:m.rsc[id].summary|default:id.title as caption %}
```
ddeboer: Why `m.rsc[id].summary` but `id.title`?

DorienD: Good question, changed this to the same as title

PR #550, opened 2019-03-01
[base] Handle missing category field as it's required

rl-king:
Without this check the response code will be a 500 with no information about the
actual problem whatsoever.

We now return a 400 and log an error with the problem.

### DIFF
--- a/modules/mod_ginger_base/controllers/controller_rest_resources.erl
+++ b/modules/mod_ginger_base/controllers/controller_rest_resources.erl
@@ -99,19 +99,14 @@ process_post(Req, State = #state{mode = collection}) ->
     %% Create resource
     {Body, Req1} = wrq:req_body(Req),
     Data = jsx:decode(Body, [return_maps, {labels, atom}]),
-    case maps:get(category, Data, undefined) of
-        undefined ->
-            lager:error("Category is missing from the request body", []),
-            {{halt, 400}, Req1, State};
-        _Cat ->
-            Props = lists:foldl(fun post_props/2, [], maps:to_list(Data)),
-            {ok, Id} = m_rsc:insert(Props, Context),
+    Props = lists:foldl(fun post_props/2, [], maps:to_list(Data)),
+    try m_rsc:insert(Props, Context) of
+        {ok, Id} ->
             %% Create edges
             lists:foreach(
               fun (Edge) ->
                       {ok, PredicateId} = m_rsc:name_to_id(maps:get(predicate, Edge), Context),
                       {ok, _EdgeId} = m_edge:insert(Id, PredicateId, maps:get(object, Edge), Context)
-
               end,
               maps:get(edges, Data, [])
              ),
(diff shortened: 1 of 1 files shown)


modules/mod_ginger_base/controllers/controller_rest_resources.erl
```diff
@@ -99,22 +99,28 @@ process_post(Req, State = #state{mode = collection}) ->
...
-      fun (Edge) ->
-              {ok, PredicateId} = m_rsc:name_to_id(maps:get(predicate, Edge), Context),
-              {ok, _EdgeId} = m_edge:insert(Id, PredicateId, maps:get(object, Edge), Context)
+    case maps:get(category, Data, undefined) of
+        undefined ->
+            lager:error("Category is missing from the request body", []),
+            {{halt, 400}, Req1, State};
+        _Cat ->
+            Props = lists:foldl(fun post_props/2, [], maps:to_list(Data)),
+            {ok, Id} = m_rsc:insert(Props, Context),
```
ddeboer: This can fail for other reasons, too (e.g. ACL). What do you think putting this in a `case` statement and returning a `400` if Zotonic produces an error when inserting the resource? The advantage is that leave responsibility about validating the insertion request to `m_rsc` instead of adding our own validation layer on top.

rl-king: This will crash with the missing `category` so a `case of` won't work unfortunately. 

I _does_ sound like a better solution though.

ddeboer: In the case you’d need to catch the exception:

```erlang
try m_rsc:insert(…) of
    {ok, …} ->
        …
catch 
    throw:{{error, no category}, _} ->
        …
    %% don’t match other exceptions to just crash on them, as the current code already did
end.
```

rl-king: I've updated the insert.

When matching `{error, eacces}` the server still returns a 500 though, not sure why. 

Also, please check the final case, not sure if I should end with

`_ ->`

or something else

modules/mod_ginger_base/controllers/controller_rest_resources.erl
```diff
@@ -100,21 +100,30 @@ process_post(Req, State = #state{mode = collection}) ->
...
+            %% Done
+            {{halt, 201}, Req2, State}
+    catch
+        throw:{{error, nocategory}, _} ->
+            lager:error("Category is missing from the request body", []),
+            {{halt, 400}, Req1, State};
+        throw:{{error, eacces}, _} ->
+            {{halt, 403}, Req1, State};
+        throw:{_} ->
+            {{halt, 500}, Req1, State}
```
ddeboer: If you leave out this case the server crashes and so returns a 500 anyway. So just remove it. 😉

rl-king:
#581

PR #552, opened 2019-03-01
Cleanup auth templates.

mworrell:
Cleanup logon templates, remove duplication between Zotonic and Ginger.

Also remove admin_seo.tpl, as 0.47 has the option to fix seo noindex flag.

Depends on Zotonic 0.47.0


DorienD:
Looking good!

mworrell:
Is there something wrong with the CI setup?

https://travis-ci.org/driebit/ginger/builds/501933941#L898

```
09:21:18.007 [error] CRASH REPORT Process 'z_pivot_rsc$testsandboxdb' with 0 neighbours exited
with reason: call to undefined function z_context:new(testsandboxdb) in gen_server:terminate/7 line 812```

**Comment by mworrell** [comment]
I have added a script `scripts/msgmerge.sh` to merge all generated .pot files into the nl.po files.
Also added an git-ignore rule for the .pot template files.

Didn't commit the result of the merge script. Looking at the results it might be better to have someone else look at the correct translations.

File: .gitignore [hunk 1]
```diff
@@ -6,6 +6,7 @@ sites/*
 !sites/.gitkeep
 modules/*
 !modules/mod_ginger*
+modules/*/translations/template
```
ddeboer: Not necessary because we already ignore `*.pot`

modules/mod_ginger_edit/templates/logon_error/message.passcode.tpl
```diff
@@ -0,0 +1,3 @@
+<p>
+    {_ The two-factor authentication passcode did not match. Please check your entry and try again. _}
+</p>
```
ddeboer: Isn’t this template one already in Zotonic? I see some other duplicates as well. Is that just to remove the logon_reminder link?

scripts/msgmerge.sh
```diff
@@ -0,0 +1,18 @@
...
+
+cd `dirname $0`/..
+for i in modules/mod_ginger*/translations/template/*.pot
+do
+    pushd `dirname $i` > /dev/null
+    if [ -e "../nl.po" ]
+    then
+        echo "Merging into nl.po from '$i'"
+        msgmerge ../nl.po `basename $i` > ../x.po
+        mv ../x.po ../nl.po
```
ddeboer: Why not have `../nl.po` as the direct target of `msgmerge` instead of moving?

scripts/msgmerge.sh
```diff
@@ -0,0 +1,18 @@
...
+    else
+        echo "Making new nl.po from '$i'"
+        cp `basename $i` ../nl.po
+    fi
+    popd > /dev/null
+done
```
ddeboer: Useful!

scripts/msgmerge.sh
mworrell: msgmerge doesn't make a new file, so we will be overwriting the input file while we are reading it.

modules/mod_ginger_edit/templates/logon_error/message.passcode.tpl
mworrell: Yes, because of the link.  Some messages have also textual variations.

.gitignore
mworrell: Without it I did see the template directories in the "files to add".
Maybe the `!modules/...` influences the other rules.

modules/mod_ginger_edit/templates/logon_error/message.passcode.tpl
ddeboer: I can’t reproduce this. As a result of `mod_translation:generate/1`? In which directory?

modules/mod_ginger_edit/templates/logon_error/message.pw.tpl
```diff
@@ -0,0 +1,3 @@
+<p>
+    {_ You entered an unknown username or password.  Please try again. _}
```
ddeboer: Should be a single space.

modules/mod_ginger_edit/templates/logon_error/message.passcode.tpl
mworrell: I assume you refer to the other discussion?

.gitignore
mworrell: I could see requests to merge the pot files, could also be dependent on the git version.

mworrell: This is what I get if I remove this line from the .gitignore:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	modules/mod_ginger_admin/translations/template/
	modules/mod_ginger_auth/translations/template/
	modules/mod_ginger_base/translations/template/
	modules/mod_ginger_edit/translations/template/
	modules/mod_ginger_foundation/translations/template/
	modules/mod_ginger_rdf/translations/template/
	modules/mod_ginger_remark/translations/template/
	modules/mod_ginger_search/translations/template/
```

modules/mod_ginger_edit/templates/logon_error/message.pw.tpl
mworrell: Removed (came from previous ginger text)

scripts/msgmerge.sh
ddeboer: You can use `msgmerge -U`.

.gitignore
ddeboer: Ah, that’s because your `msgmerge.sh` creates *.po files in those directories (which are not ignored).

mworrell: Ah, it did create some garbage in those directories with an older version.  Removed the .gitignore rule.

modules/mod_ginger_edit/templates/logon_error/message.tpl
```diff
@@ -0,0 +1,3 @@
+{# Generic logon error message, use "logon_error/message.<reason>.tpl for specific error templates #}
+
+<p>{_ Something went wrong, please try again later. _}</p>
```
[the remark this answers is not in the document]
mworrell: We might get some errors like timeouts, they should be mapped to this generic error.

scripts/msgmerge.sh
mworrell: Using `--update` now

ddeboer:
Thanks!

PR #557, opened 2019-03-08
[base] Patch resource via REST api

rl-king:
Since there is no PATCH method we use post.

Posting at `/data/resources/id` will update the resource at `id`


ddeboer:
Actually, updating a single REST resource shouldn’t be PATCH but PUT. Does PUT work?

rl-king:
It it supported, what's implemented here updates the resource, not replace it

ddeboer:
Of course, you’re right! `m_rsc:update/3` indeed _partially_ updates a resource. Still, I’d argue that PUT is semantically closer here than POST, which is about adding new resources.

rl-king:
K I'll update it

PR #558, opened 2019-03-11
Add reset password controller

linuss:


modules/mod_ginger_base/controllers/controller_auth.erl
```diff
@@ -48,6 +50,39 @@ process_post(Req, State = #state{mode = new}) ->
     {ok, Id} = mod_signup:signup([{email, Email}], [{identity, Identity}], RequestConfirm, Context),
     ok = mod_signup:request_verification(Id, Context),
     {{halt, 204}, Req1, State};
+process_post(Req, State = #state{mode = reset_password}) ->
+    Context = State#state.context,
+    {Body, Req1} = wrq:req_body(Req),
+    #{u := Username,
```
ddeboer: `u` is a tad cryptic; why not just `username`?

modules/mod_ginger_base/controllers/controller_auth.erl
```diff
@@ -48,6 +50,39 @@ process_post(Req, State = #state{mode = new}) ->
...
+      secret := Secret,
+      password1 := Password1,
+      password2 := Password2} = jsx:decode(Body, [return_maps, {labels, atom}]),
+    PasswordMinLength = z_convert:to_integer(
+                          m_config:get_value(mod_ginger_base, password_min_length, "6", Context)),
+    case {Password1,Password2} of
+        {A,_} when length(A) < PasswordMinLength ->
```
ddeboer: We should check against `m.config.mod_admin_identity.password_regex`, too. (Zotonic checks against this in templates.)

modules/mod_ginger_base/controllers/controller_auth.erl
linuss: You're right, I'll adapt it. As for the why, I've basically copied and adapted this code from Zotonic's `mod_authentication`, which uses this shorthand :)

ddeboer: Not this part, though. 😉

modules/mod_ginger_base/controllers/controller_auth.erl
```diff
@@ -48,6 +50,46 @@ process_post(Req, State = #state{mode = new}) ->
...
+    {Body, Req1} = wrq:req_body(Req),
+    #{username := Username,
+      secret := Secret,
+      password1 := Password1,
+      password2 := Password2} = jsx:decode(Body, [return_maps, {labels, atom}]),
+    PasswordMinLength = z_convert:to_integer(
+                          m_config:get_value(mod_ginger_base, password_min_length, "6", Context)),
+    %% Default to ".", which matches on any character
+    PasswordRegex = z_convert:to_list(
+                      m_config:get_value(mod_admin_identity, password_regex, ".", Context)),
```
ddeboer: It would be a little more efficient to execute to skip `re:run` when no `password_regex` is defined.

linuss: Fixed :)

rl-king:
I think `validate/2` should be renamed to something like `chain`.

- I don't expect it to have side effects but it does.
- It looks like a generic api to validate something but it isn't
- A validator can have any return value which means it will break if the order of the list is changed

linuss:
Yeah, `validate` might not be completely accurate. I'll rename it to `pipeline`, would that work for you? :)

PR #560, opened 2019-03-12
Support the new connect-dialog

mworrell:
This adds support for the new combined connect dialog

Depends on https://github.com/zotonic/zotonic/pull/2019

The new Zotonic connect-dialog combines the following tabs:

 * Find
 * New
 * Upload

It also support showing a preview of found pages in an overlay.


DorienD:
I don't think we want this new modal standard in the ginger edit, but that we can choose per site to use it.

mworrell:
@DorienD Instead of adding configuration options it might be best to add a separate module with a different variation of the dialog. Then you can make it exactly as you want and Ginger itself can leave the connect dialog as-is, which removes dependencies on the core admin dialogs.

PR #567, opened 2019-03-21
Improvement ginger rest

linuss:


rl-king:
@linuss This closes https://github.com/driebit/ginger/pull/550 right?

linuss:
@rl-king Hm, well, yes and no. It will log and return the proper error, but it won't return the right status code (all exceptions still result in a 500). I'd put the check on category in a `malformed_request` callback

PR #572, opened 2019-03-29
[base] Add Html-video-embed to media

rl-king:


modules/mod_ginger_base/models/m_ginger_rest.erl
```diff
@@ -134,7 +134,19 @@ with_media(Rsc = #{<<"id">> := Id}, Mediaclasses, Context) ->
...
-                    end
+                    end;
+                <<"text/html-video-embed">> ->
+                    case proplists:get_value(video_embed_code, Medium) of
+                        undefined ->
+                            Rsc;
+                        EmbedCode ->
+                            Url = EmbedCode,
+                            Height = proplists:get_value(preview_height, Medium, null),
+                            Width = proplists:get_value(preview_width, Medium, null),
```
DorienD: Can we add the video ID in here? Might come in handy when you don't want to use the iframe code but for example the youtube api or vimeo api.

rl-king: I haven't used the youtube or vimeo api in an Elm app yet. Also not entirely sure how we would 
as well, probably through a `custom element`. Also, in theory the REST api isn't only for Elm apps of course. 

If we include the ID we'd also need to include the provider (youtube, vimeo, ect).

Any examples of where the iframe doesn't suffice?

PR #578, opened 2019-04-15
Fix a typo in the `wrq:set_resp_body` function call

linuss:
We should realllllyyy write some tests...


modules/mod_ginger_base/controllers/controller_auth.erl
```diff
@@ -81,7 +81,7 @@ process_post(Req, State = #state{mode = reset_password}) ->
             m_identity:delete_by_type(UserId, "logon_reminder_secret", Context),
             {{halt, 204}, Req1, State};
         {error, Error} ->
-            {{halt, 400}, wrg:set_resp_body(z_convert:to_list(Error), Req1), State}
+            {{halt, 400}, wrq:set_resp_body(z_convert:to_list(Error), Req1), State}
```
ddeboer: 😱

Issue #566, opened 2019-03-21
[base] resource_exists in REST api doesn't return 404 but 500 


rl-king:
Calling `/data/resources/:id` with a non existing id returns a 500 instead of 404.

https://github.com/driebit/ginger/blob/master/modules/mod_ginger_base/controllers/controller_rest_resources.erl#L60

ddeboer:
This is caused by `controller_rest` accepting a map instead of a list for `State`. Webzmachine doesn’t handle a map argument correctly.

rl-king:
Ah thanks, interesting. Quite a refactor to change that as well.

rl-king:
As suggested, replacing maps with proplists in the dispatch rules fixes this issue

PR #579, opened 2019-04-19
[base] Use propslist in REST dispatch, add is_authorized

rl-king:
fixes: #566

### DIFF
--- a/modules/mod_ginger_base/controllers/controller_rest_edges.erl
+++ b/modules/mod_ginger_base/controllers/controller_rest_edges.erl
@@ -1,6 +1,7 @@
 -module(controller_rest_edges).
 
 -export([ init/1
+        , is_authorized/2
         , service_available/2
         , allowed_methods/2
         , resource_exists/2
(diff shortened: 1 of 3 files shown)


modules/mod_ginger_base/controllers/controller_rest_edges.erl
```diff
@@ -24,14 +25,28 @@
...
+    case lists:member(wrq:method(Req), ['POST', 'DELETE']) of
+        false ->
+            {true, Req, State};
+        true ->
+            Context = State#state.context,
+            case z_acl:user(Context) of
+                undefined ->
+                    {{halt, 401}, Req, Context};
+                _Id ->
+                    {true, Req, State}
```
robvandenbogaard: Instead of just true, we'd probably want to have an authorization function taking the user id here.

rl-king: To further check if the action is allowed?

robvandenbogaard: Yes, either by calling into the ginger acl rules or by calling a function that implements a rest specific rulebase (e.g. "can only change own data" or something).

robvandenbogaard: Ah.. the zotonic functions for manipulating edges already have acl checks, so indeed your suggestion of adding tests while leaving this `true` is a good idea.

modules/mod_ginger_base/controllers/controller_rest_edges.erl
```diff
@@ -24,14 +25,28 @@
...
+init(Args) ->
+    Mode = proplists:get_value(mode, Args),
     {ok, #state{mode = Mode}}.
 
 service_available(Req, State) ->
     Context = z_context:continue_session(z_context:new(Req, ?MODULE)),
     {true, Req, State#state{context = Context}}.
 
+is_authorized(Req, State) ->
+    case lists:member(wrq:method(Req), ['POST', 'DELETE']) of
```
linuss: Took me too long to get this, maybe add a comment?

modules/mod_ginger_base/controllers/controller_rest_edges.erl
```diff
@@ -24,14 +25,29 @@
...
+init(Args) ->
+    Mode = proplists:get_value(mode, Args),
     {ok, #state{mode = Mode}}.
 
 service_available(Req, State) ->
     Context = z_context:continue_session(z_context:new(Req, ?MODULE)),
     {true, Req, State#state{context = Context}}.
 
+is_authorized(Req, State) ->
+    %% Don't check HEAD
```
linuss: %% Auth checks are only necessary for these methods

PR #585, opened 2019-04-26
[base] Add string contains function

rl-king:
Let's start making a collection of functions we keep rewriting every project cause they're missing in the erlang std libs.

`g_*` Seems like a good naming scheme filled with puns.

### DIFF
--- a/modules/mod_ginger_base/support/g_string.erl
+++ b/modules/mod_ginger_base/support/g_string.erl
@@ -16,6 +16,8 @@
 
 %% @doc Check if a string contains another string
 -spec contains(list(), list()) -> boolean().
+contains([], []) ->
+    true;
 contains(_, []) ->
     false;
 contains(Pred, Str) ->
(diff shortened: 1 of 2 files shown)


modules/mod_ginger_base/support/g_string.erl
```diff
@@ -0,0 +1,34 @@
...
+%%% API
+%%%===================================================================
+
+%% @doc Check if a string contains another string
+-spec contains(list(), list()) -> boolean().
+contains([], []) ->
+    true;
+contains(_, []) ->
+    false;
+contains(Pred, Str) ->
```
DorienD: Change `Pred` to something more clear

robvandenbogaard: Yeah, I'd propose `Pattern`.

rl-king:
Let's not put this in Ginger but in a seperate repo/package

PR #593, opened 2019-05-09
Dirty fix to (try and) fetch at least "limit" results when searching

linuss:


modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -38,19 +38,31 @@ to_json(Req, State) ->
...
             _ ->
-                Result = z_search:search({Type, arguments(Req)}, {Offset + 1, Limit}, Context),
-                #{ result => [ search_result(R, Context) || R <- Result#search_result.result
-                                                                , is_visible(R, Context)
-                             ]
-                 , total => Result#search_result.total
+                %% Because is_visible cannot be added to the ElasticSearch query, we retrieve 10 times
+                %% the number of results we'll be returning, filter them on is_visible, and then
+                %% return the first `limit` results, to make (reasonably) sure we're returning the
+                %% right amount of results
```
robvandenbogaard: How would you determine the correct offset for getting the next page of results?

PR #594, opened 2019-05-14
Add the 'JSX_FORCE_MAPS` envvar to the Dockerfile

linuss:


CyBeRoni:
Typically, for layering, it's better to add env vars that you don't need to influence on the line with make. See the DEBUG=1 bit.

PR #596, opened 2019-05-20
Remove unnecessary 'isPrimaryTopic' filter from DBpedia Sparql query

linuss:
This filter prevented us from seeing useful data, and the check for the wiki
link is done in the templates as well, so is unnecessary here.


ddeboer:
Thanks!

PR #605, opened 2019-06-07
 [Admin, Base] Add theme colors to different environments

loetie:
To quickly distinguish if a user is on acceptance or production (or developers on dev and test)


modules/mod_ginger_admin/package.json
```diff
@@ -10,16 +10,18 @@
...
-        "gulp-uglify": "^3.0.0"
+        "gulp-babel": "^7.0.0",
+        "gulp-css-globbing": "^0.2.2",
+        "gulp-livereload": "^4.0.1",
+        "gulp-plumber": "^1.2.0",
+        "gulp-postcss": "^8.0.0",
+        "gulp-rename": "^1.4.0",
+        "gulp-sass": "^4.0.2",
+        "gulp-uglify": "^3.0.1",
+        "lost": "^8.3.0"
```
ddeboer: These changes seem unrelated. Were they intended as a clean-up or update?

mworrell:
For discerning the live/testing/staging, maybe we could add a colored bar at the top or one of the sides?  Another option is to add a 'theme' class to the admin, so that we always echo that configured class name in the body tag (within the admin). Then you can add some css to change colors and/or some texts.

ddeboer:
To make the environment (i.e. theme) selection more generic, we could base it on a server configuration variable (in `zotonic.config`).

mworrell:
Yes, that is a good idea. At least something that can completely overruled by the server configuration.

We can then fetch it using `m.site.environment`.

In this way it is still possible to have a not-yet-live site on a production server, and switch it to `production` in a very visible way.

mworrell:
In the code I see the following possible stages:

 * `dev`
 * `test`
 * `acceptance`
 * `production`

Checking DTAP (https://en.wikipedia.org/wiki/Development,_testing,_acceptance_and_production) I see:

 * Development
 * Test
 * Acceptance
 * Production
 * Education (for learning the environment by new users)
 * Backup (a backup site for disaster recovery)

Shall we use these names for possible values?

mworrell:
Proposal:

 1. Add `zotonic.config` setting `environment` (possible values, see above)
 2. Add optional site configuration `environment` (overrules 1, so is possible to have a backup or education site on production server)
 3. Add `m.site.environment` which echos the current environment, defaults to `production`
 4. Set a class `environment-{{ m.site.environment }}` on the html tag in admin and other Zotonic base templates (idem for Ginger base templates)
 5. Set a HTTP response header: `X-DTAP-Environment: ...` reflecting the current environment.

mworrell:
I added an issue for Zotonic here: https://github.com/zotonic/zotonic/issues/2120

mworrell:
See the  merge request here: https://github.com/zotonic/zotonic/pull/2137

mworrell:
The merge request in Zotonic is awaiting acceptance.

modules/mod_ginger_admin/package.json
DorienD: For the new gulp setup.

mworrell:
Closing this - Zotonic has these environment colors now.

PR #608, opened 2019-07-02
Fix total of search results

linuss:


modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -34,15 +34,13 @@ to_json(Req, State) ->
...
                  , facets => facets(Result#search_result.facets)
                  };
             _ ->
                 Result = z_search:search({Type, arguments(Req)}, {Offset + 1, Limit}, Context),
-                #{ result => [ search_result(R, Context) || R <- Result#search_result.result
-                                                                , is_visible(R, Context)
-                             ]
-                 , total => Result#search_result.total
+                #{ result => total(Result, Context)
+                 , total => results(Result, Context)
```
ddeboer: Should this be the other way around?

PR #617, opened 2019-08-20
[base] Escape html in Ginger Rest API

rl-king:
Don't unescape translated texts

### DIFF
--- a/modules/mod_ginger_base/models/m_ginger_rest.erl
+++ b/modules/mod_ginger_base/models/m_ginger_rest.erl
@@ -74,11 +74,17 @@ with_edges(Rsc = #{<<"id">> := Id}, Predicates, Context) ->
 %% @doc Get resource translations.
 -spec translations(atom() | {trans, proplists:proplist()}, z:context()) -> translations().
 translations({trans, Translations}, Context) ->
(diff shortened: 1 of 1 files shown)


modules/mod_ginger_base/models/m_ginger_rest.erl
```diff
@@ -74,11 +74,17 @@ with_edges(Rsc = #{<<"id">> := Id}, Predicates, Context) ->
...
 translations(undefined, Context) ->
     [{z_trans:default_language(Context), <<"">>}];
 translations(Value, Context) ->
-    [{z_trans:default_language(Context), z_html:unescape(filter_show_media:show_media(Value, Context))}].
+    [{z_trans:default_language(Context), show_media(Value, Context)}].
+
+show_media(Value, Context) ->
+    case filter_show_media:show_media(Value, Context) of
+        Result when is_binary(Result) -> Result;
+        Result when is_list(Result) -> << B || B <- lists:flatten(Result) >>
```
robvandenbogaard: This list is not a list of diverse translations?

rl-king: No `filter_show_media:show_media` returns a list of binaries

PR #618, opened 2019-08-20
Add command to Makefile for db-imports in non-docker environments

elizabethprins:


Makefile
```diff
@@ -24,6 +24,7 @@ help:
 	@echo "  gulp site=your_site     Run Gulp in a site directory"
 	@echo "  clean-node              Delete all node_modules directories"
 	@echo "  import-db-file          Import database from file in ginger data dir (site=site-name file=site-dump.sql)"
+	@echo "  import-db-file-alt      Import database from file in ginger data dir (site=site-name file=data/site-dump.sql)"
```
robvandenbogaard: Maybe we need to name this `import-db-file-dockerless`?

Makefile
```diff
@@ -15,27 +15,28 @@ export ERLASTIC_SEARCH_JSON_MODULE := jsx
...
+	@echo "  addsite name=site-name     Create a new site"
+	@echo "  api-doc                    Generate API doc"
+	@echo "  deps                       Install dependencies"
+	@echo "  disco                      Make your site available at http://[sitename].[username].ginger.test"
+	@echo "  down                       Stop containers"
+	@echo "  dump-db site=site-name     Dump database to /data directory using pg_dump"
+	@echo "  gulp site=your_site        Run Gulp in a site directory"
+	@echo "  clean-node                 Delete all node_modules directories"
+	@echo "  import-db-file             Import database from file in ginger data dir (site=site-name file=site-dump.sql)"
+	@echo "  import-db-file-dockerless  Import database from file in ginger data dir (site=site-name file=data/site-dump.sql)"
```
DorienD: Why change this to `data/site-dump.sql` and not keep it the same as import-db-file?

PR #620, opened 2019-08-23
[Base][Foundation] Refactor video for size and alignment

DorienD:


modules/mod_ginger_base/templates/media/_caption.tpl
```diff
@@ -0,0 +1,8 @@
+{% if caption or id.o.actor %}
+    <figcaption>
+        {% if caption == "-" %}{% else %}
+            <p>{{ caption }}{% if m.rsc[id].o.author %} {_ By: _} <a href="{{ m.rsc[m.rsc[id].o.author[1]].page_url }}">{{ m.rsc[m.rsc[id].o.author[1]].title }}</a>{% endif %}</p>
```
robvandenbogaard: Would `id.o.author` be okay instead and in line with the style in the 1st line?

DorienD: Good one!

DorienD: I refactored this.

PR #623, opened 2019-09-05
[base] Convert search api query param type

rl-king:


rl-king:
Lists won't be possible anymore. Providing multiple of the same will result in an AND query. 

It will indeed crash on named refs 🤔

PR #632, opened 2019-10-25
[SPA] Mod spa improvements

rl-king:
* Fix language attribute (should be lowercase)
* Remove a11y unfriendly browser scaling
* Render site menus as js objects and pass on to elm as flags
* Add opengraph and twitter card meta tags
* Add charset to properly render ligatures and the like
* Add base.tpl, which loads some zotonic js files needed for public admin pages
* Add mediaclass.config

### DIFF
--- a/modules/mod_ginger_spa/templates/base.tpl
+++ b/modules/mod_ginger_spa/templates/base.tpl
@@ -19,7 +19,7 @@
 
 #################################
 //-->
-<html lang="{{ q.lang|default:"nl" }}">
+<html lang="{{ q.lang|default:"nl"|escape }}">
     <head>
         <meta charset="utf-8" />
         <title>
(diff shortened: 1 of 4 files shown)


rl-king:
Not sure if the mediaclass.config should be in here, could also be put the skel.

modules/mod_ginger_spa/templates/base.tpl
```diff
@@ -0,0 +1,55 @@
...
+
+        Oudezijds Voorburgwal 282
+                1012 GL Amsterdam
+                   020 - 420 8449
+                  info@driebit.nl
+                   www.driebit.nl
+
+#################################
+//-->
+<html lang="{{ q.lang|default:"nl" }}">
```
elizabethprins: `{{ q.lang|default:"nl"|escape }}`

PR #655, opened 2020-01-23
[Edit] Change translation to something nicer

DorienD:


EdwardAbboud (approved):
As much as i like the passive aggressive version, this is a better change :D

PR #658, opened 2020-02-12
[rdf] Fix JSON-LD type serialization

ddeboer:
We used to produce JSON-LD documents like:

```
"@type": {
    "@id": "http://..."
}
```

However, this doesn't validate against https://json-ld.org/playground/, so change it to:

```
"@type": ["http://..."]
```

which does validate.


rl-king:
@ddeboer We're about to tag a new release and while reviewing the changes we were wondering what impact of this change is to existing code. Won't all code expecting the existence of "@id" break?

ddeboer:
Hey @rl-king! You’re right, this is definitely a BC break. In practice:

- internal callers of the affected `ginger_json_ld:serialize_to_map/1` are mainly `controller_rdf:rdf/2` and `m_rdf:m_find_value/3`; they serialize straight to JSON (which still works fine)
- any external JSON-LD parsers were unable to parse `@type` before anyway because it was not conforming to [the JSON-LD standard](https://w3c.github.io/json-ld-syntax/#specifying-the-type).

rl-king:
👍 thanks for the quick response!

PR #663, opened 2020-04-03
[Base] Refactor jquery selection on ID

DorienD:


DorienD:
> There always only one element with this id?

yes

PR #664, opened 2020-04-03
Always filter resources that have is_unfindable=true

rl-king:
Added to the search api endpoint

The resources endpoint continues to return resources that have `is_unfindable` set to `true`

### DIFF
--- a/modules/mod_ginger_base/controllers/controller_search.erl
+++ b/modules/mod_ginger_base/controllers/controller_search.erl
@@ -61,8 +61,11 @@ params(Req) ->
 -spec arguments(wrq:rd()) -> proplists:proplist().
 arguments(Req) ->
     RequestArgs = wrq:req_qs(Req),
+    Default = {filter, ["is_unfindable", '<>', true]},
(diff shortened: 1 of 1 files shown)


modules/mod_ginger_base/controllers/controller_search.erl
```diff
@@ -61,8 +61,11 @@ params(Req) ->
 -spec arguments(wrq:rd()) -> proplists:proplist().
 arguments(Req) ->
     RequestArgs = wrq:req_qs(Req),
+    Default = {filter, ["is_unfindable", '<>', true]},
```
robvandenbogaard, line 64: Does this mean we *keep* items unfindable or that we want to *exclude* them? It seems the latter from the name "FilteredArguments" (although not unambiguously so), however it seems not entirely obvious the surrounding logic will use them as such...
I would say we only want to keep items that have `unfindable == false` meaning `findable == true`.

robvandenbogaard: [After a slack conversation with RLK] Ah, ok it is to indicate what is begin filtered out.

PR #668, opened 2020-04-22
[Base] Add filter to get current env

DorienD:
A filter to get the current environment to use in the templates.


rl-king:
Shouldn't this be implemented as a model? @loetie @robvandenbogaard

rl-king (commented):
Note sure how to remove myself from the reviewers list, but by doing this my review "todo" list in GH is empty :)

mworrell:
In Zotonic we have `m.site.environment`, I propose to use that.

Issue #670, opened 2020-07-07
[base + modules] Remove side effects from manage_schema


robvandenbogaard:
To avoid concurrency problems in loading data models between dependent modules, remove (database) side effects from `manage_schema/2` functions, because in its current form the intended use is to return the data model declaration and let Zotonic take care of installing/updating in dependency order.

PR #671, opened 2020-07-07
[base + modules] Issue #670 - Remove side effects in manage_schema functions

robvandenbogaard:


rl-king:
ping @mworrell

modules/mod_ginger_base/mod_ginger_base.erl
```diff
@@ -168,9 +168,7 @@ manage_schema(_Version, Context) ->
                 ]}
             ]}
         ]
-    },
-    z_datamodel:manage(?MODULE, Datamodel, Context),
-    schema:create_identity_if_not_exists(editor_dev, "redacteur", "redacteur", Context).
```
rl-king, line 173: Where did this go?

modules/mod_ginger_banner/mod_ginger_banner.erl
```diff
@@ -14,12 +14,11 @@
 ]).
 
 manage_schema(_Version, Context) ->
```
rl-king, line 16: Context -> _Context (unused)

modules/mod_ginger_embed/mod_ginger_embed.erl
```diff
@@ -21,7 +21,7 @@
 -include("zotonic.hrl").
 
 manage_schema(_Version, Context) ->
```
rl-king, line 23: Context -> _Context (unused)

modules/mod_ginger_message/mod_ginger_message.erl
```diff
@@ -16,7 +16,7 @@
 ]).
 
 manage_schema(install, Context) ->
-    Datamodel = #datamodel{
+    #datamodel{
```
rl-king, line 19: Context -> _Context (unused)

modules/mod_ginger_base/mod_ginger_base.erl
robvandenbogaard, line 173: Proposal to remove the automatic creation of `editor_dev`. If we need one we should provide it in fixtures.

modules/mod_ginger_tagger/mod_ginger_tagger.erl
```diff
@@ -19,18 +19,20 @@
 ]).
 
 manage_schema(install, Context) ->
```
rl-king, line 21: Version is `install`?

Context -> _Context (unused)

modules/mod_ginger_base/mod_ginger_base.erl
rl-king: Ok, I've never used it, maybe @loetie has

modules/mod_ginger_rdf/mod_ginger_rdf.erl
```diff
@@ -34,18 +34,8 @@
...
-        resources=[
-            {rdf_content_group, content_group, [
-                {title, <<"RDF resources">>}
-            ]}
-        ]
-    },
-    z_datamodel:manage(?MODULE, Datamodel, Context),
-
+    %% TODO: Legacy code ahead! This function must not have side effects;
+    %% it should only return the data model.
```
robvandenbogaard, line 38: We should look into this case to see if it is still necessary to keep this here, or if it could be in a better place.

modules/mod_ginger_base/mod_ginger_base.erl
robvandenbogaard: It went nowhere, as I think it is not used / we should not use it (here) and rethink it when we need it.

modules/mod_ginger_tagger/mod_ginger_tagger.erl
mworrell: Version type signature is:

```
-type manage_schema() :: install
                       | {upgrade, integer()}.
```

modules/mod_ginger_base/mod_ginger_base.erl
mworrell: There is also an `manage_data/2` function. That is called _after_ the `manage_schema/2` has been handled.

See https://test.zotonic.com/page/1353/modules

```erlang
manage_data(_Version, Context) ->
    %% Whatever data needs to be installed after the datamodel
    %% has been installed.
    ok.
```

mworrell: Use `manage_data/2`, it is called after `manage_schema/2`

```erlang
manage_data(_Version, Context) ->
    %% Whatever data needs to be installed after the datamodel
    %% has been installed.
    ok.
```

modules/mod_ginger_rdf/mod_ginger_rdf.erl
mworrell: Use `manage_data/2` to update/install data after the schema has been installed.

modules/mod_ginger_base/mod_ginger_base.erl
robvandenbogaard: Good point, thanks.

mworrell: BTW `schema` is a dangerous name for a module. I can imagine that other projects will also try to use this generic name.  Better prefix it with `ginger_`, as it is a Ginger-only module.

robvandenbogaard: Yes, I completely agree, that struck me too.

PR #675, opened 2020-08-18
Ginger REST specify edges

emiflake:
Built on top of fix for #670

The API using the erlang parser is mostly just to get it usable right away.

### DIFF
--- a/modules/mod_ginger_banner/mod_ginger_banner.erl
+++ b/modules/mod_ginger_banner/mod_ginger_banner.erl
@@ -13,7 +13,7 @@
     manage_schema/2
 ]).
 
-manage_schema(_Version, Context) ->
+manage_schema(_Version, _Context) ->
     #datamodel{
         resources = [
             {message_banner, text, [
--- a/modules/mod_ginger_banner/mod_ginger_banner.erl
+++ b/modules/mod_ginger_banner/mod_ginger_banner.erl
@@ -14,12 +14,11 @@
 ]).
 
 manage_schema(_Version, Context) ->
-    Datamodel = #datamodel{
+    #datamodel{
         resources = [
             {message_banner, text, [
                 {title, "Banner"},
                 {is_published, false}
             ]}
         ]
-    },
-    z_datamodel:manage(?MODULE, Datamodel, Context).
+    }.
--- a/modules/mod_ginger_base/controllers/controller_rest_resources.erl
+++ b/modules/mod_ginger_base/controllers/controller_rest_resources.erl
@@ -103,8 +103,9 @@ to_json(Req, State = #state{mode = collection}) ->
                    wrq:req_qs(Req)
                   )
                 ),
+        EdgePath = parse_path(proplists:get_value("edge_path", wrq:req_qs(Req), "")),
         Ids = z_search:query_(Args, Context),
-        Json = jsx:encode([rsc(Id, Context, true) || Id <- Ids, m_rsc:is_visible(Id, Context)]),
+        Json = jsx:encode([with_deep_edges(m_ginger_rest:rsc(Id, Context), EdgePath, Context) || Id <- Ids, m_rsc:is_visible(Id, Context)]),
         {Json, Req, State}
     catch
         _:Error ->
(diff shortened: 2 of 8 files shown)


modules/mod_ginger_banner/mod_ginger_banner.erl
```diff
@@ -13,13 +13,12 @@
...
+    #datamodel{
         resources = [
             {message_banner, text, [
                 {title, "Banner"},
                 {is_published, false}
             ]}
         ]
-    },
-    z_datamodel:manage(?MODULE, Datamodel, Context).
+    }.
```
robvandenbogaard, line 24: Yes, correct - manage_schema is called to return a model only; the caller will materialize it.

modules/mod_ginger_base/controllers/controller_rest_resources.erl
```diff
@@ -103,8 +103,9 @@ to_json(Req, State = #state{mode = collection}) ->
                    wrq:req_qs(Req)
                   )
                 ),
+        EdgePath = parse_path(proplists:get_value("edge_path", wrq:req_qs(Req), "")),
```
robvandenbogaard, line 106: Does this mean one can only traverse/include a single edge path in a single resource request? I would think we need to be able to specify multiple paths.

modules/mod_ginger_rdf/mod_ginger_rdf.erl
```diff
@@ -58,7 +48,17 @@ manage_schema(_, Context) ->
                 Context
```
robvandenbogaard, line 48: Yes indeed, the predicate update above is a tricky sideeffect.

modules/mod_ginger_base/controllers/controller_rest_resources.erl
```diff
@@ -225,15 +228,67 @@ trans({Key, Value}, Acc) ->
...
+                    #rsc_list{list = Objects} = m_rsc:o(Id, Predicate, Context),
+                    [
+                     #{
+                      <<"predicate_name">> => Predicate,
+                      <<"resource">> =>
+                          with_deep_edges(
+                            m_ginger_rest:rsc(Object, Context),
+                            DeeperPredicates,
+                            Context)
+                     } || Object <- Objects
```
robvandenbogaard, line 285: Ah we don't need the `is_visible` here, because `m_rsc:o` is covering that?

modules/mod_ginger_base/controllers/controller_rest_resources.erl
emiflake: Currently we gain two abilities with this new query parameter:
- Specify depth of edges
- Specify a filter for the edges at each depth
Together I called this an 'edge path'.

But indeed, we could want to resolve multiple 'edge paths'. I think naively, we could solve this by just taking a list of 'edge paths' and running `with_deep_edges` for each of the paths, but of course we would be wasting effort by traveling the same edges twice.

modules/mod_ginger_rdf/mod_ginger_rdf.erl
emiflake: These changes are yours! (I built this PR on top of that branch, and it took these two commits with it)

modules/mod_ginger_base/controllers/controller_rest_resources.erl
emiflake: Good question, I have a hard time tracing back if that is the case. This code was partly inherited from [m_ginger_rest.erl's with_edges/3](https://github.com/driebit/ginger/blob/master/modules/mod_ginger_base/models/m_ginger_rest.erl#L58) which uses `m_rsc:o` instead of `m_edge:get_edges` in `with_edges/2`.

modules/mod_ginger_base/controllers/controller_rest_resources.erl
robvandenbogaard: Yes, maybe good to start out with this and if we see we need more we can easily extend.

modules/mod_ginger_rdf/mod_ginger_rdf.erl
robvandenbogaard: Oh haha! :D Now you say it..I recall doing something similar indeed :joy:

modules/mod_ginger_base/controllers/controller_rest_resources.erl
robvandenbogaard: I agree. I'm having a hard time finding that check in `m_rsc:o` too. It seems to just forward to `m_edge:get_edges`. So we actually should check visibility here explicitly too.

robvandenbogaard: Great!

rl-king:
moved to #676

PR #677, opened 2021-02-19
Change og image tags to https

M-Skipemski:


robvandenbogaard (approved):
In a next update we also need to move the fallback image into the files/archive directory to be able to use image_url in the else block too.

PR #678, opened 2021-05-04
Dev053 - Template render controller

mworrell:
Add template rendering to mod_ginger_spa

### DIFF
--- a/modules/mod_ginger_base/support/ginger_environment.erl
+++ b/modules/mod_ginger_base/support/ginger_environment.erl
@@ -14,13 +14,20 @@
 %% @doc Get current environment
 -spec get(#context{}) -> dev|acc|prod.
 get(Context) ->
-    case is_dev(Context) of
-        true -> dev;
-        false ->
-            case is_acc(Context) of
-                true -> acc;
-                false -> prod
-            end
+    case z_config:get(environment) of
+        production ->
+            case is_dev(Context) of
+                true -> dev;
+                false ->
+                    case is_acc(Context) of
+                        true -> acc;
+                        false -> prod
+                    end
+            end;
+        acceptance ->
+            acc;
+        _ ->
+            dev
     end.
 
 is_dev(Context) ->
--- a/modules/mod_ginger_spa/README.md
+++ b/modules/mod_ginger_spa/README.md
@@ -1,7 +1,42 @@
 mod_ginger_spa
-========================
+==============
 
 Default template and dispatch rules for a single page app.
 
 _Only routes for which a Ginger resource with the same path exists will be served by the `page.tpl` template.
 Other routes should be manually added to the sites dispatch rules._
+
+
+Template rendering
+------------------
+
+The `controller_ginger_spa_template` renders templates and returns the result as plain text.
+
+Pass the template name in the path, example:
+
+    https://example.com/render-template/public/test.tpl
+
+This renders the template `"public/test.tpl"`.
+
+Pass arguments via the query string:
+
+    https://example.com/render-template/public/test.tpl?id=1
+
+Or in a POST body as `application/x-www-form-urlencoded` or `multipart/form-data`.
+
+Use the argument `is_cat_include=1` to render the template using a *catinclude*.
+Any `id` argument is always added as `id` to the template arguments (after mapping using `m_rsc:rid/2`), in
+this way the `id` is always available and usable for the optional catinclude.
+
+Security
+........
+
+As in Zotonic 0.x template models are not access checked there is a security using the name of the template.
+
+Templates starting with `public/...` can be rendered by anonymous users.
+
+Templates starting with `member/...` can be rendered by authenticated users.
+
+Administrators are allowed to render all templates.
+
+
--- a/modules/mod_ginger_spa/controllers/controller_ginger_spa_template.erl
+++ b/modules/mod_ginger_spa/controllers/controller_ginger_spa_template.erl
@@ -1,0 +1,69 @@
+%% @copyright Driebit BV 2021
+%% @doc Controller to render templates. Post arguments via the query arguments.
+
+-module(controller_ginger_spa_template).
+
+-export([init/1, service_available/2, charsets_provided/2, content_types_provided/2]).
+-export([allowed_methods/2, resource_exists/2]).
+-export([process_post/2, provide_content/2]).
+
+-include_lib("controller_webmachine_helper.hrl").
+-include_lib("include/zotonic.hrl").
+
+init(DispatchArgs) -> {ok, DispatchArgs}.
+
+service_available(ReqData, DispatchArgs) when is_list(DispatchArgs) ->
+    Context1 = z_context:new_request(ReqData, DispatchArgs, ?MODULE),
+    ?WM_REPLY(true, Context1).
+
+allowed_methods(ReqData, Context) ->
+    {[ 'POST', 'GET', 'HEAD' ], ReqData, Context}.
+
+charsets_provided(ReqData, Context) ->
+    {[{"utf-8", fun(X) -> X end}], ReqData, Context}.
+
+content_types_provided(ReqData, Context) ->
+    {[{"text/plain", provide_content}], ReqData, Context}.
+
+resource_exists(ReqData, Context) ->
+    Context1 = ?WM_REQ(ReqData, Context),
+    Context2 = z_context:ensure_all(Context1),
+    ?WM_REPLY(true, Context2).
+
+process_post(ReqData, Context) ->
+    Context1 = ?WM_REQ(ReqData, Context),
+    Result = render(Context1),
+    ReqData1 = wrq:set_resp_body(Result, ReqData),
+    {{halt, 200}, ReqData1, Context}.
+
+provide_content(ReqData, Context) ->
+    Context1 = ?WM_REQ(ReqData, Context),
+    Result = render(Context1),
+    ?WM_REPLY(Result, Context1).
+
+render(Context) ->
+    Template = z_context:get_q("*", Context),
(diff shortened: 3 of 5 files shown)


modules/mod_ginger_spa/dispatch/dispatch
```diff
@@ -1,3 +1,6 @@
 [{home, [], controller_template, [ {template, "page.tpl"}, {id, home} ]},
- {search, ["search"], controller_template, [ {template, "page.tpl"} ]}
+ {search, ["search"], controller_template, [ {template, "page.tpl"} ]},
+
+ % Render a template the path is the template, pass arguments via the query post or get
+ {spa_render, ["render-template", '*'], controller_ginger_spa_template, []}
```
robvandenbogaard, line 5: Should the url have `api` in it? Like `example.com/api/template/render`? To avoid potential clashes with project domain paths.

modules/mod_ginger_spa/controllers/controller_ginger_spa_template.erl
```diff
@@ -0,0 +1,69 @@
...
+            <<>>
+    end.
+
+is_catinclude(Context) ->
+    z_convert:to_bool(z_context:get_q("catinclude", Context)).
+
+is_viewable("/" ++ _, _Context) -> false;
+is_viewable("public/" ++ _, _Context) -> true;
+is_viewable("member/" ++ _, Context) -> z_auth:is_auth(Context);
+is_viewable(T, Context) when is_list(T) -> z_acl:is_admin(Context).
```
robvandenbogaard, line 69: I think I understand why we need an extra authorisation layer here; because templates can have side-effects.

Possibly this template rendering is only effective and beneficial if we can force it to be immutable data-wise, and transparent to all templates. Part of the efficiency lies in being able to reuse existing templates.

A good alternative mechanism (also to immutability) might be to enforce that the template is reachable by url (full pages) _or_ it resides in a `parts` or `partials` dir (I'd prefer those over `public` because it might not be public).

Probably any module should be able to have a `partials` dir, and its use could be restricted by the module's _usage acl rule_. Maybe we can implement (per module) authorisation by allowing acl rules for partials to be added to the acl list in admin?

mworrell: The _parts_ / _partials_ is ok with me.

In general templates don't have and shouldn't have side effects.
But... there are some admin templates that show config values, those are the ones that need to be protected.
In the 1.x we solved this with access control on all models, which we can't add to the 0.x

We can check the module permissions, as we could do a template lookup and then check the `use` permission on the module the template belongs to.

This is not fool-proof, but can come quite far.

modules/mod_ginger_spa/README.md
```diff
@@ -1,7 +1,42 @@
...
+Security
+........
+
+As in Zotonic 0.x template models are not access checked there is a security using the name of the template.
+
+Templates starting with `public/...` can be rendered by anonymous users.
+
+Templates starting with `member/...` can be rendered by authenticated users.
+
+Administrators are allowed to render all templates.
```
robvandenbogaard, line 40: I was wondering why we need this convention: searching back in our slack conversations I realised this is a way to avoid having to implement per template access rules, or acl via module config.

modules/mod_ginger_spa/templates/public/test.tpl
```diff
@@ -0,0 +1,2 @@
+{% debug %}
+{% print q.v %}
```
robvandenbogaard, line 2: Maybe this should be less dangerous (-looking) default content, without `debug`? I Imagine you'll easily forget it is there - wait..maybe if we change the default dispatch rule to explicitly reference test.tpl, you'd be bound to remove or fix it in an actual project.

modules/mod_ginger_spa/dispatch/dispatch
```diff
@@ -1,3 +1,6 @@
 [{home, [], controller_template, [ {template, "page.tpl"}, {id, home} ]},
- {search, ["search"], controller_template, [ {template, "page.tpl"} ]}
+ {search, ["search"], controller_template, [ {template, "page.tpl"} ]},
+
+ % Render a template the path is the template, pass arguments via the query post or get
+ {spa_render, ["api", "render-template", "public", '*' ], controller_ginger_spa_template, []}
```
robvandenbogaard, line 5: She earlier comment on public/test.tpl, maybe we should make the default spa_render rule specific, for discoverability of the default test template.

modules/mod_ginger_base/support/ginger_environment.erl
```diff
@@ -14,13 +14,20 @@
...
+                false ->
+                    case is_acc(Context) of
+                        true -> acc;
+                        false -> prod
+                    end
+            end;
+        acceptance ->
+            acc;
+        _ ->
+            dev
```
robvandenbogaard, line 30: Is this change extra or needed for the template render controller?

modules/mod_ginger_rdf/models/m_rdf_export.erl
```diff
@@ -23,7 +23,10 @@ to_rdf(Id, Context) ->
 %%      - media.
 -spec to_rdf(m_rsc:resource(), [module()], z:context()) -> m_rdf:rdf_resource().
 to_rdf(Id, Ontologies, Context) ->
-    Properties = m_rsc:get_visible(Id, Context),
+    Properties = case m_rsc:get_visible(Id, Context) of
+        undefined -> [];
+        Props -> Props
+    end,
```
robvandenbogaard, line 29: Just for the record, from our slack conversation: This rdf change was snucked in, unrelated to template render controller. Maybe we need to apply this in a separate pull request; maybe it has been lying around for too long because it potentially fixes bugs?

modules/mod_ginger_spa/controllers/controller_ginger_spa_template.erl
```diff
@@ -0,0 +1,83 @@
...
+content_types_provided(ReqData, Context) ->
+    {[{"text/plain", provide_content}], ReqData, Context}.
+
+resource_exists(ReqData, Context) ->
+    Context1 = ?WM_REQ(ReqData, Context),
+    Context2 = z_context:ensure_all(Context1),
+    ?WM_REPLY(true, Context2).
+
+process_post(ReqData, Context) ->
+    Context1 = ?WM_REQ(ReqData, Context),
```
robvandenbogaard, line 34: Is this strictly necessary? Would returning Context2 in resource_exists (which is visited before process_post) not already have the ReqData WM_REQ'ed in?

modules/mod_ginger_spa/controllers/controller_ginger_spa_template.erl
```diff
@@ -0,0 +1,83 @@
...
+                {id, m_rsc:rid(z_context:get_q("id", Context), Context)}
+            ],
+            Template1 = case is_catinclude(Context) of
+                true -> {cat, Template};
+                false -> Template
+            end,
+            {Data, _} = z_template:render_to_iolist(Template1, Vars, Context),
+            iolist_to_binary(Data);
+        {error, _} ->
+            lager:info("[~p] Denied render of template \"~s\"", [ z_context:site(Context), m_req:get(path, Context) ]),
```
robvandenbogaard, line 58: Is denied the right wording? If the template is in the url, any raised error would be caused by that there is no matching dispatch rule. Note that a template name/path given in the `template` context property is not checked for this. Is this as intended?

modules/mod_ginger_spa/controllers/controller_ginger_spa_template.erl
```diff
@@ -0,0 +1,83 @@
...
+    case z_context:get(template, Context) of
+        undefined ->
+            case path(Context) of
+                "api/render-template/" ++ Template ->
+                    {ok, drop_slash(Template)};
+                _ ->
+                    {error, enoent}
+            end;
+        Template ->
+            {ok, Template}
```
robvandenbogaard, line 75: It doesn't do the dispatch check here - is that as intended, or a loophole?

robvandenbogaard (commented):
Looks good; can be merged after discussing the comments and maybe good to try it out using elm while still on the branch.

modules/mod_ginger_base/support/ginger_environment.erl
mworrell: This smuggled in - for compatibility between the Ginger environment and the Zotonic environment.

modules/mod_ginger_spa/templates/public/test.tpl
mworrell: The debug just echos what is passed, and afaik all data is escaped.

modules/mod_ginger_spa/README.md
mworrell: Indeed - the problem is that the models in 0.x do not have access control on their value functions.

This is not needed in the 1.x, as there all models have full access control.

modules/mod_ginger_rdf/models/m_rdf_export.erl
mworrell: It potentially fixes bugs.

Ideally it should have been a separate merge request.

modules/mod_ginger_spa/controllers/controller_ginger_spa_template.erl
mworrell: Might be anything, but I think this is good enough.

modules/mod_ginger_spa/controllers/controller_ginger_spa_template.erl
mworrell: yes, then it is hard coded in the dispatch rule, so the responsibility of the person programming that dispatch rule.

PR #684, opened 2022-01-24
[build] Move CI to GitHub Actions

ddeboer:


.github/workflows/test.yml
```diff
@@ -0,0 +1,55 @@
...
+
+            -   name: Compile
+                working-directory: ${{ env.ZOTONIC_DIR }}
+                run: make all
+
+            -   name: Run tests
+                working-directory: ${{ env.ZOTONIC_DIR }}
+                env:
+                    ERL: "erl -noshell"
+                run: $GITHUB_WORKSPACE/scripts/runtests.sh
```
robvandenbogaard, line 55: I didn't work with github workflows yet, so this is a nice introduction. Clear script!

robvandenbogaard: The Slack notification was removed, is that intentional/checked with Marco? Maybe it is indeed not needed anymore - though I am used to see them signalling successful tests..are these the ones sent to the room ginger-notifications?

robvandenbogaard (commented):
Good to merge if the removal of notifications was confirmed as ok.

.github/workflows/test.yml
ddeboer: @robvandenbogaard The Slack notification is only saying whether the tests failed/succeeded, so it’s really a non-essential addition to the feedback from GitHub itself. Because of this, I’d suggest to forget about the notification for now.

If you want to [re-add the notification](https://axolo.co/blog/p/top-4-github-action-slack-integration), that would require some configuration on the Driebit Slack (which I have no admin access to). So first let’s see if it’s really needed, and only if so, set up the Driebit Slack.

robvandenbogaard: I now see that we probably did not have these notifications for some time, so it is ok to remove it (was confusing the notifications about test/acceptance getting updated).

robvandenbogaard (approved):

PR #687, opened 2022-01-27
[rdf] Fix nested properties

ddeboer:


robvandenbogaard (approved):

modules/mod_ginger_rdf/tests/mod_ginger_rdf_tests.erl
```diff
@@ -156,14 +156,26 @@ deserialize_test() ->
...
         {title, {trans, [
             {nl, <<"Een mooie titel">>}
         ]}},
         {is_published, true}
     ],
+    AuthorProps = [
+        {category, person},
+        {title, {trans, [
+            {nl, <<"Auteur van het artikel">>}
+        ]}},
```
robvandenbogaard, line 173: Reminded by that a name is normally not translated (but that is not the point here I know) I am wondering if it would be good to also check what happens with untranslated values?

ddeboer: Good idea! Made the property monolingual so we’re now testing the two variants.

robvandenbogaard (approved):

PR #691, opened 2022-02-04
[rdf] Make explicit that triple object can be a resource

ddeboer:
This is used for nested resources that have blank nodes.

### DIFF
--- a/modules/mod_ginger_rdf/include/rdf.hrl
+++ b/modules/mod_ginger_rdf/include/rdf.hrl
@@ -19,6 +19,11 @@
     type = undefined :: undefined | ginger_uri:uri()
 }).
 
+-record(rdf_resource, {
+    id :: ginger_uri:uri(),
+    triples = [] :: [m_rdf:triple()]
+}).
+
 -record(triple, {
     %% DEPRECATED: type property is deprecated. Construct an object = #rdf_value{value = ...}
     %% instead.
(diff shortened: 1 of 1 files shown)


modules/mod_ginger_rdf/include/rdf.hrl
```diff
@@ -27,15 +32,10 @@
     subject :: undefined | binary(),
     subject_props = [] :: proplists:proplist(),
     predicate :: m_rdf:predicate(),
-    object :: ginger_uri:uri() | #rdf_value{},
+    object :: ginger_uri:uri() | #rdf_value{} | #rdf_resource{},
```
robvandenbogaard, line 35: Is the difference between ginger_uri:uri() and #rdf_resource{} that the latter has its triples inlined (as a kind of embedded document)? I wonder if it complicates things if these seemingly overlapping variants are both included?

robvandenbogaard: Seeing the comment above on blank nodes, indeed the difference is inlined triples. Never mind my remark.

robvandenbogaard (approved):

modules/mod_ginger_rdf/include/rdf.hrl
ddeboer: Correct, this is about inline triples. I do agree with you that our data model could be improved with better support for blank nodes by introducing (for example) a `blank_node_identifier()` type next to `uri()`, but that is somewhat out of scope of this simple PR solely intended to document the current behaviour better.

PR #692, opened 2022-02-16
mod_ginger_rdf: add caching of dbpedia lookups in the tkvstore.

mworrell:
This fixes an issue where keyword metadata is not show if nl.dbpedia.org is down.

### DIFF
--- a/modules/mod_ginger_rdf/models/m_dbpedia.erl
+++ b/modules/mod_ginger_rdf/models/m_dbpedia.erl
@@ -16,8 +16,11 @@
     is_wikidata_uri/1
 ]).
 
+-define(CACHE_TYPE, <<"m_dbpedia:uri:fetch">>).
+
+
 %% @doc Usage: m.dbpedia["http://nl.dbpedia.org/resource/Nederland"]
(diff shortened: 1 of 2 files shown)


robvandenbogaard (approved):

modules/mod_ginger_rdf/models/m_dbpedia.erl
```diff
@@ -46,14 +49,66 @@ get_resource(<<"nl.dbpedia.org", _/binary>> = Uri, Context) ->
...
-        Context
-    ).
+            get_resource_cached(Uri, Language, Context).
+    % z_depcache:memo(
+    %     fun() ->
+    %         get_resource_cached(Uri, Language, Context)
+    %     end,
+    %     {Uri, Language},
+    %     Context
+    % ).
```
ddeboer, line 61: @mworrell Are you using tkvstore because that is backed by Postgres so will persist between restarts? Or are there other reasons to prefer tkvstore over depcache?

mworrell: Indeed, because of long term caching.

Especially after the months-long problem with nl.dbpedia.org

PR #693, opened 2022-02-18
[rdf] Serialize queries as Hydra collections

ddeboer:
* Add a controller at `/hydra/123` that serializes Zotonic queries
  as Hydra collections, either in JSON-LD or Turtle.
* Hydra gives machine-readable pagination over large collections.
  See﻿ https://www.hydra-cg.com/spec/latest/core/.
* Filter out empty triple objects in both JSON-LD and Turtle serializations.
  This makes it easy to dump some triples that may not have a value,
  such as previous or next page in our case. 
* The `/hydra` endpoint can be used as the RDF distribution of datasets,
  which are a subcategory of `query`. 
* Add documentation.

### DIFF
--- a/modules/mod_ginger_rdf/controllers/controller_hydra.erl
+++ b/modules/mod_ginger_rdf/controllers/controller_hydra.erl
@@ -1,0 +1,154 @@
+%% @doc Serve Hydra representation of a collection resource or one of its subcategories.
+%% @see https://www.hydra-cg.com/spec/latest/core/
+-module(controller_hydra).
+-author("David de Boer <david@ddeboer.nl>").
+
+-export([
+    init/1,
+    service_available/2,
+    resource_exists/2,
+    content_types_provided/2,
+    to_json_ld/2,
+    to_turtle/2
+]).
+
+-include_lib("controller_webmachine_helper.hrl").
+-include_lib("zotonic.hrl").
+-include_lib("../include/rdf.hrl").
+
+-record(state, {
+    ontology :: atom(),
+    limit :: non_neg_integer(),
+    context :: z:context()
+}).
+
+init(Args) ->
+    {ok, Args}.
+
+service_available(ReqData, Args) ->
+    Context = z_context:new_request(ReqData, [], ?MODULE),
+    Context2 = cors_headers:allow_all_cors(Context),
+    State = #state{
+        context = Context2,
+        ontology = proplists:get_value(ontology, Args),
+        limit = proplists:get_value(limit, Args, 100)
+    },
+    case wrq:method(ReqData) of
+        'OPTIONS' ->
+            {{halt, 204}, z_context:get_reqdata(Context), State};
+        _ ->
+            {true, z_context:get_reqdata(Context), State}
+    end.
+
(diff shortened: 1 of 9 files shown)


modules/mod_ginger_rdf/controllers/controller_hydra.erl
```diff
@@ -0,0 +1,154 @@
...
+    undefined;
+page(_, 0) ->
+    undefined;
+page(QueryParams, undefined) ->
+    proplists:delete("page", QueryParams);
+page(QueryParams, PageNumber) when is_number(PageNumber) andalso PageNumber >= 1 ->
+    [{"page", PageNumber} | proplists:delete("page", QueryParams)].
+
+id(Context) ->
+    m_rsc:rid(z_context:get_q(id, Context), Context).
```
robvandenbogaard, line 154: Maybe if this is named `get_rid_from_q_id` or something (`resource_id_from_q`?), it would have been clearer for me what was happening in the calling code above. Not too important, because here it is very clear what it does.

modules/mod_ginger_rdf/dispatch/dispatch
```diff
@@ -1,4 +1,5 @@
 [
     {rsc_json_ld, ["rdf", id], controller_rdf, [{serialization, json_ld}, {ontology, schema_org}]},
-    {rsc_turtle, ["turtle", id], controller_rdf, [{serialization, turtle}, {ontology, schema_org}]}
+    {rsc_turtle, ["turtle", id], controller_rdf, [{serialization, turtle}, {ontology, schema_org}]},
+    {hydra, ["hydra", id], controller_hydra, [{ontology, schema_org}]}
```
robvandenbogaard, line 4: If one would like to be able to include multiple ontologies (which might be completely out of scope for the current use case), would making the ontology parameter a list complicate things too much?

modules/mod_ginger_rdf/support/cors_headers.erl
```diff
@@ -0,0 +1,24 @@
...
+                V ->
+                    z_context:set_resp_header(atom_to_list(K), V, Acc)
+            end
+        end,
+        Context,
+        [
+            {'Access-Control-Allow-Origin', "*"},
+            {'Access-Control-Max-Age', undefined}
+        ]
+    ).
```
robvandenbogaard, line 24: The comment says that this function does not look at configuration, but the implementation seems to do so. Also if no config value is present, no header is changed/set - is this the intended behaviour?

ddeboer: Fixed in #695.

modules/mod_ginger_rdf/dispatch/dispatch
ddeboer: Done in #697.

modules/mod_ginger_rdf/controllers/controller_hydra.erl
ddeboer: Yeah, that was my (lazy) consideration: because the function body is so small, it should be quite clear what it does. Still, you make a valid point to have function names a tad more expressive than merely `id`.

PR #694, opened 2022-02-21
[rdf] Return nested subject as URI

ddeboer:
Because the RDF resource's id is always and URI, it makes sense to
return a JSON-LD `@id` rather than a string.


robvandenbogaard (commented):
Would you like to change the `owl:sameAs` predicate to something less semantically distracting? (Sorry for getting trapped into thinking that the specific predicate `sameAs` was part of the test - the test context is that this can be anything; like `firstPage` or `somePredicate` ;))

ddeboer:
Done!

robvandenbogaard (approved):

PR #697, opened 2022-02-21
[rdf] Support multiple ontologies in controllers

ddeboer:
* Keep `ontology` for BC with existing dispatch rules in sites.

### DIFF
--- a/modules/mod_ginger_rdf/controllers/controller_hydra.erl
+++ b/modules/mod_ginger_rdf/controllers/controller_hydra.erl
@@ -17,7 +17,7 @@
 -include_lib("../include/rdf.hrl").
 
 -record(state, {
-    ontology :: atom(),
+    ontologies :: atom(),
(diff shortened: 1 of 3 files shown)


modules/mod_ginger_rdf/controllers/controller_hydra.erl
```diff
@@ -17,7 +17,7 @@
 -include_lib("../include/rdf.hrl").
 
 -record(state, {
-    ontology :: atom(),
+    ontologies :: atom(),
```
robvandenbogaard, line 20: Should the type specifier be a list of atoms instead?

ddeboer: Nice catch! Adding those types without actually having Dialyzer on CI is of course more for illustrative/documentary purposes, but still, it should be correct. 😄

robvandenbogaard (approved):

PR #698, opened 2022-02-21
[rdf] Include only visible resources in Hydra output

ddeboer:


modules/mod_ginger_rdf/controllers/controller_hydra.erl
```diff
@@ -76,8 +78,8 @@ to_hydra(ReqData, State) ->
         total = Total,
         next = Next,
         prev = Previous
-    } = z_search:search_pager({query, [{query_id, QueryId}]}, Page, Limit, Context),
-    RdfResults = [m_rdf_export:to_rdf(Id, [State#state.ontology], Context) || Id <- Result],
+    } = z_search:search_pager({State#state.query_type, search_arguments(QueryId, Context)}, Page, Limit, Context),
```
robvandenbogaard, line 81: I wonder if the `query_type` should come from the resource - is it a parameter of the hydra request (similar to the ontology parameter)? I think the query type is inseparable from the collection resource because it could result in a completely different collection of data if the type is changed.

modules/mod_ginger_rdf/controllers/controller_hydra.erl
```diff
@@ -152,3 +154,14 @@ page(QueryParams, PageNumber) when is_number(PageNumber) andalso PageNumber >= 1
...
+%% @doc If the resource has a query text, use that to search.
+%%      If not, fall back to the resource's outgoing haspart edges.
+-spec search_arguments(m_rsc:resource(), z:context()) -> proplists:proplist().
+search_arguments(Id, Context) ->
+    case m_rsc:p(Id, query, Context) of
+        <<>> ->
+            [{is_published, true}, {hassubject, [Id, haspart]}];
+        _QueryText ->
+            [{is_published, true}, {query_id, Id}]
+    end.
```
robvandenbogaard, line 167: For collections containing both query and haspart links, this would result in omitting haspart data - is this a compromise/as intended?

modules/mod_ginger_rdf/controllers/controller_hydra.erl
ddeboer: The `query_type` comes from the dispatch rule (just like `ontologies` does). 

> I think the query type is inseparable from the collection resource because it could result in a completely different collection of data if the type is changed.

True, but we have (e.g.) `ginger_search` for Ginger-specific search criteria, which may be used in the query. Anyway, I could hard-code this to `query` but decided that parameterising in the dispatch rule is a bit better.

modules/mod_ginger_rdf/controllers/controller_hydra.erl
ddeboer: Definitely a compromise that could be improved upon. However, search pagination will be incorrect if we simply add the two sets of resources together (those from the dynamic query and the haspart objects).

It would be more elegant to combine the search criteria to say: give me all resources `({query_id, Id} OR {hassubject, [Id, haspart]}) AND is_published`. However, I’ve found no way to express this in a Zotonic query.

robvandenbogaard: Yes indeed, agreeed.

robvandenbogaard (approved):

PR #699, opened 2022-03-07
[base] Include unique name in /data/resources output

robvandenbogaard:
As the resource unique name is a core concept and tool within Zotonic/Ginger domain modelling it is quite surprising that it wasn't included in the resource representations yet. Did I miss anything that opposes inclusion of the resource unique name?


robvandenbogaard:
As the resource unique name is a core concept and tool within Zotonic/Ginger domain modelling it is quite surprising that it wasn't included in the resource representations yet. Did I miss anything that opposes inclusion of the resource unique name?

robvandenbogaard:
Note from Emma explaining why we could do without so far:
In projects where we need `name` we configure it as a custom property `name => ginger_string` in the `types` parameter of the Zotonic `config`.

mworrell (approved):

PR #702, opened 2022-06-17
Add mod_elasticsearch2 support for Elasticsearch v7+

mworrell:
This must be used with `mod_elasticsearch2`

### DIFF
--- a/modules/mod_ginger_adlib_elasticsearch2/mod_ginger_adlib_elasticsearch2.erl
+++ b/modules/mod_ginger_adlib_elasticsearch2/mod_ginger_adlib_elasticsearch2.erl
@@ -1,0 +1,70 @@
+%% @doc Zotonic/Adlib integration
+-module(mod_ginger_adlib_elasticsearch2).
+-author("Driebit <tech@driebit.nl>").
+
+-mod_title("Adlib in ElasticSearch").
+-mod_prio(500).
+-mod_description("Makes Adlib data searchable by indexing it in ElasticSearch (v7+)").
+-mod_depends([mod_elasticsearch2, mod_ginger_adlib]).
+-mod_schema(1).
+
+-export([
+    init/1,
+    index/1,
+    types/1,
+    manage_schema/2,
+    observe_adlib_update/2
+]).
+
+-include_lib("mod_ginger_adlib/include/ginger_adlib.hrl").
+-include_lib("zotonic.hrl").
+
+init(Context) ->
+    default_config(index, index(Context), Context),
+    prepare_elasticsearch_index(Context).
+
+manage_schema(_, Context) ->
+    prepare_elasticsearch_index(Context).
+
+%% @doc Create index and mappings for each type in the index.
+-spec prepare_elasticsearch_index(z:context()) -> ok.
+prepare_elasticsearch_index(Context) ->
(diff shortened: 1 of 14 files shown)


modules/mod_ginger_adlib_elasticsearch2/mod_ginger_adlib_elasticsearch2.erl
```diff
@@ -0,0 +1,70 @@
...
+%% @doc Create index and mappings for each type in the index.
+-spec prepare_elasticsearch_index(z:context()) -> ok.
+prepare_elasticsearch_index(Context) ->
+    {Version, Mapping} = ginger_adlib_elasticsearch2_index_mapping:default_mapping(Context),
+    {ok, _} = elasticsearch2_index:upgrade(index(Context), Mapping, Version, Context),
+    ok.
+
+observe_adlib_update(#adlib_update{date = _Date, database = Database, record = #{<<"@attributes">> := #{<<"priref">> := Priref}} = Record}, Context) ->
+    lager:debug("Indexing Adlib record ~s from database ~s", [Priref, Database]),
+    MappedRecord = ginger_adlib_elasticsearch2_mapper:map(Record, ginger_adlib_elasticsearch_mapping),
```
robvandenbogaard, line 38: Is it ok that the module name passed is without the `2` suffix?

mworrell: Good one - that should have the `2`

modules/mod_ginger_collection/mod_ginger_collection.erl
```diff
@@ -4,36 +4,25 @@
 -mod_title("Linked data collections").
 -mod_description("Linked data and media collection view and search interface powered by Elasticsearch").
 -mod_prio(200).
--mod_depends([mod_ginger_base, mod_elasticsearch, mod_ginger_rdf]).
+-mod_depends([mod_ginger_base, elasticsearch, mod_ginger_rdf]).
```
robvandenbogaard, line 7: Should this also be `elasticsearch2` instead. Was the `mod_` prefix dropped?

robvandenbogaard: I didn't realise that ginger_collection would always be dependent on (mod) elasticsearch; I thought it was optional to use elastic.

robvandenbogaard: Aha, so this one should wait until those PRs are available. Nice, to have a generic elasticsearch dep. So usage in code here should also be e.g. `elasticsearch:put..` without the 2?

mworrell: We can't add the generic `elasticsearch` module, as that would give an Erlang module naming problem with installs where we have both elasticsearch and elasticsearch2.

mworrell: It would be possible to add that using notifiers though, which is our preferred way of decoupling.

robvandenbogaard: Ah yes; thought so, that I was mixing up deps management and module namespaces.

robvandenbogaard (approved):

PR #716, opened 2022-10-21
Elasticsearch2 Adlib support

mworrell:
Add support for mod_elasticsearch2 to the Adlib/collection modules.

In ES7+ the _type is removed, as we used it for the Adlib database reference when storing documents we now have to combine the type and priref of a record to make an unique document `_id`  (as multiple databases can have the same priref).

The filter `collection_object_url` can now be used to make `collection_object` urls, this instead of manually combining the type and id in the template (or Erlang code).

See also: https://github.com/driebit/mod_elasticsearch2/pull/5


cdfa (approved):
LGTM! Handig die `collection_object_url`

PR #722, opened 2022-12-13
Update erfgoed brabant search filters

cdfa:


robvandenbogaard (approved):
Note that this project specific code actually should be factored out (by using a notification/observer, or passing a whitelist function). We can do that later in a cleanup operation.

PR #724, opened 2022-12-15
Foundation module update

cdfa:
Fixes for breaking changes in sass up to 1.52 for `mod_ginger_foundation`, which is often imported by ginger sites.


modules/mod_ginger_foundation/lib/css/src/blocks/_list.scss
```diff
@@ -156,23 +156,9 @@
...
-    &:hover,
-    &:active,
-    &:focus {
-        @extend %btn--primary:hover;
-    }
-
-    + .list__more {
-        margin-top: 20px;
-        background: lighten($mainThemeColor, 20);
-    }
```
robvandenbogaard, line 175: Note from review with Colin: .list__more should stay. And the border-radius of %btn (18px) is different from $interfaceRadius (4px).

modules/mod_ginger_foundation/lib/css/__default-screen.scss
```diff
@@ -1,98 +1,22 @@
...
-@import '../../../../../modules/mod_ginger_foundation/lib/css/src/base';
-@import '../../../../../modules/mod_ginger_foundation/lib/css/src/icons';
-@import '../../../../../modules/mod_ginger_foundation/lib/css/src/buttons';
-@import '../../../../../modules/mod_ginger_foundation/lib/css/src/buttons';
+@import '/lib/css/src/variables';
+@import '/lib/css/src/mixins';
+@import '/lib/css/src/base';
+@import '/lib/css/src/icons';
+@import '/lib/css/src/buttons';
+@import '/lib/css/src/buttons';
```
robvandenbogaard, line 7: Note from review with Colin: We are curious about how abs/rel paths work here (What is the effect of using the leading slash?); maybe to avoid any confusion we can consistently use relative paths.

DorienD: So this file `__default-screen.scss` was used as an example for sites that want to use foundation but not everything. And we needed to use the absolute paths because of the makefile or something. I guess because we stopped working with 'globbing'?

robvandenbogaard: The reason to stop globbing is because it was not supported anymore in newer sass versions.
Aha, thanks for the clarification, we suspected it to be meant as an example indeed 👍

robvandenbogaard (approved):

PR #726, opened 2022-12-22
Fix an issue with dynamic elastic-ids

cdfa:
Looking up an object with `m.collection_object["Object"][id]` was broken when id was an elastic2 id (with `::Type` appended). This commit fixes this and allows lookup with `m.collection_object[id]` as well.


mworrell:
Ah yes, bit hustling with the typed ids.  As they are probably getting mixed up in templates.

mworrell (approved):

PR #730, opened 2023-02-28
Fix a problem where nl.wikipedia.org gives timeouts with http (but not https)

mworrell:
As the 'http' site gives timeouts, and it will redirect to 'https' anyway, force metadata lookups to use the 'https' url.

### DIFF
--- a/modules/mod_ginger_rdf/support/dbpedia.erl
+++ b/modules/mod_ginger_rdf/support/dbpedia.erl
@@ -75,7 +75,7 @@ get_resource(Uri, Language) ->
 %% @doc The nl dbpedia is missing the thumbnails of entries. Fetch the thumbnail from
 %% the wikipedia page.
 get_resource_thumbnail(Triples) ->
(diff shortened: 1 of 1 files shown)


modules/mod_ginger_rdf/support/dbpedia.erl
```diff
@@ -75,16 +75,21 @@ get_resource(Uri, Language) ->
...
-        undefined ->
-            undefined;
-        WikipediaUrl ->
-            case z_url_metadata:fetch(WikipediaUrl) of
-                {ok, MD} ->
-                    z_url_metadata:p(image, MD);
-                {error, _} ->
-                    undefined
-            end
+    case ?DEBUG(fetch_wikipedia_url(Triples)) of
```
robvandenbogaard, line 78: Is it correct that there should be a DEBUG here?

mworrell: oops :-)

robvandenbogaard (approved):

PR #737, opened 2024-07-01
Replace polyfill.io with fastly's alternative

pasqu4le:
See the issue: [OEI-128](https://driebit.atlassian.net/browse/OEI-128)

For reference, see: https://sansec.io/research/polyfill-supply-chain-attack

### DIFF
--- a/modules/mod_ginger_base/templates/head/head.tpl
+++ b/modules/mod_ginger_base/templates/head/head.tpl
@@ -27,10 +27,6 @@
         "css/logon.css"
     %}
 
-    {% include
-        "head/polyfill.tpl"
-    %}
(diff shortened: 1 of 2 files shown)


robvandenbogaard:
Isn't it better to refrain from directly linking to 3rd party libraries in this way? This moves the attack surface from a Chinese supplier to probably a party susceptible to surveillance from other sides ;)

pasqu4le:
You are right, I didn't do it right away because I wasn't sure if we needed the service at all.

It turns out that we can remove it entirely, so I updated this PR now.

robvandenbogaard (approved):

PR #738, opened 2024-07-05
[mod_ginger_rdf] Make parser more robust against variants of nesting

robvandenbogaard:
- When in a json+ld response the @context is top-level and multiple subjects are described in the @graph part, pass on the context into the recursion. Also, if a local @context is found in the description of a single subject, let it override (merge with) the global context, as expected.
- Let the caller of deserialize specify an id where the call was done for - so if multiple subjects are described we can pick the right one for top level.

### DIFF
--- a/modules/mod_ginger_rdf/support/ginger_json_ld.erl
+++ b/modules/mod_ginger_rdf/support/ginger_json_ld.erl
@@ -239,8 +239,7 @@ compact_predicate(Predicate) ->
 %% @doc Deserialize a JSON-LD document into an RDF resource.
 -spec deserialize(tuple() | list()) -> #rdf_resource{}.
 deserialize(JsonLd) when is_map(JsonLd) ->
-    Context = maps:get(<<"@context">>, JsonLd, #{}),
-    deserialize_with_context(undefined, JsonLd, Context);
+    deserialize_with_context(undefined, JsonLd, #{});
 deserialize(JsonLd) ->
     %% Fall back to mochijson {struct, ...} structure
     open(JsonLd).
@@ -248,8 +247,7 @@ deserialize(JsonLd) ->
 %% Guide the parser; indicate that we are actually looking for a specific id
 %% so we don't get confused when we encounter multiple subjects in the jsonld.
 deserialize(Id, JsonLd) when is_map(JsonLd) ->
-    Context = maps:get(<<"@context">>, JsonLd, #{}),
-    deserialize_with_context(Id, JsonLd, Context).
+    deserialize_with_context(Id, JsonLd, #{}).
 
 deserialize_with_context(Id, JsonLd, Context) ->
     LocalContext = maps:get(<<"@context">>, JsonLd, #{}),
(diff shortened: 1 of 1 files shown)


modules/mod_ginger_rdf/support/ginger_json_ld.erl
```diff
@@ -239,25 +240,36 @@ compact_predicate(Predicate) ->
...
+    deserialize_with_context(undefined, JsonLd, Context);
+deserialize(JsonLd) ->
+    %% Fall back to mochijson {struct, ...} structure
+    open(JsonLd).
+
+%% Guide the parser; indicate that we are actually looking for a specific id
+%% so we don't get confused when we encounter multiple subjects in the jsonld.
+deserialize(Id, JsonLd) when is_map(JsonLd) ->
+    Context = maps:get(<<"@context">>, JsonLd, #{}),
+    deserialize_with_context(Id, JsonLd, Context).
```
pasqu4le, line 252: Nitpick, but isn't this the same as:
```suggestion
deserialize(Id, JsonLd) when is_map(JsonLd) ->
    deserialize_with_context(Id, JsonLd, #{}).
```
given that `deserialize_with_context` fetches the local context too?

modules/mod_ginger_rdf/support/ginger_json_ld.erl
```diff
@@ -239,25 +240,36 @@ compact_predicate(Predicate) ->
...
+deserialize_with_context(Id, JsonLd, Context) ->
+    LocalContext = maps:get(<<"@context">>, JsonLd, #{}),
+    TotalContext = maps:merge(Context, LocalContext),
     maps:fold(
         fun(Key, Value, Acc) ->
-            deserialize(Key, Value, Acc, Context)
+            deserialize(Key, Value, Acc, TotalContext)
         end,
-        #rdf_resource{},
+        #rdf_resource{id = Id},
```
pasqu4le, line 261: This seems an unnecessary addition, I don't see this `Id` being used 🤔 
```suggestion
        #rdf_resource{},
```

pasqu4le (approved):

modules/mod_ginger_rdf/support/ginger_json_ld.erl
robvandenbogaard: This is actually the core of the fix ;) - in deserialize the id of the containing rdf_resource is not set, (because) it can be ambiguous in the case of multiple subjects (with id) described, so what we do if we know the uri of the requested resource, is set it explicitly. It might not be used in this local piece of code, but it is essential as identifier of the (top level) rdf_resource.

modules/mod_ginger_rdf/support/ginger_json_ld.erl
robvandenbogaard: Agreed, this applies to both function clauses (also the one at :241).

robvandenbogaard: Applied the suggestion to both clauses.

PR #743, opened 2024-11-06
ZZCS-33 Fix a problem where incorrect language results would be returned for dbpedia fetches

mworrell:
When fetching a resource from dbpedia, and If a predicate is present in multiple languages, then these languages are mixed in all rows of the search results.

It looks like a kind of combinatorial explosion of all language values.

Ideally the spqarql query should group by language of the returned predicates.
As that doesn't seem to be trivial, we now group the returned predicates by language and try to find the _best matching_ triple value for the requested language.

### DIFF
--- a/modules/mod_ginger_rdf/models/m_dbpedia.erl
+++ b/modules/mod_ginger_rdf/models/m_dbpedia.erl
@@ -12,6 +12,7 @@
     m_value/2,
     get_resource/2,
     get_resource/3,
+    get_resource_fetch/4,
     is_dbpedia_uri/1,
     is_wikidata_uri/1,
     task_resource_update/3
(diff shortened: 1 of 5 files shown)


modules/mod_ginger_rdf/models/m_dbpedia.erl
```diff
@@ -80,7 +84,8 @@ task_resource_update(Uri, Language, Context) ->
     end,
     ok.
 
-get_resource_fetch(Uri, Language, StaleData, Context) ->
+get_resource_fetch(Uri0, Language, StaleData, Context) ->
+    Uri = strip_protocol(Uri0),
     Key = cache_key(Uri, Language),
     case dbpedia:get_resource(<<"http://", Uri/binary>>, Language) of
```
pasqu4le, line 90: Why not use the URI with the `http://` as a  key instead?

We already have a `ginger_uri:https` function, we could add a `ginger_uri:http` which forces the protocol to be HTTP and here do something like:
```suggestion
    Uri = ginger_uri:http(Uri0),
    Key = cache_key(Uri, Language),
    case dbpedia:get_resource(Uri, Language) of
```

pasqu4le (approved):

modules/mod_ginger_rdf/models/m_dbpedia.erl
mworrell: Cleaned up.

PR #744, opened 2025-01-13
OEI-70 ginger foundation

Amber-Cloud:
Adjusted scss to be compiled without errors on both node sass and dart sass projects.


EdwardAbboud (approved):
Looks alright to me!

PR #751, opened 2025-07-11
OEI-70: add legacy mixins file

M-Skipemski:
@EdwardAbboud, since it was not possible to make a version that compiled for both the old and new sass compiler, I added a seperate file for the legacy projects like you suggested. I you encounter an old project it should be pretty easy to just make it include this file instead of the current one. Probably not needed much since you dit so many nix updates, but nice as a backup.


EdwardAbboud (approved):
Looks good!

PR #753, opened 2025-09-23
Add ACL rule for insert/delete user group

loetie:
Moderator can use mod_admin_identity for user management reasons. Part of user management is adding/removing user groups to a user. Yet not be able to change to/from managers user group.

Giving access to the whole mod_acl_user_group seems to have more consequences then just managing user groups.


M-Skipemski (approved):
Looks good

